PR #8, opened 2019-12-26
Centralize ID generation

adroitwhiz:
The code to generate random IDs was used in 3 different places--this PR centralizes the code into one function that is imported and called from those places.

This also opens up the possibility of using [the same ID generation code as Scratch](https://github.com/LLK/scratch-vm/blob/develop/src/util/uid.js).


PullJosh:
Will merge in a moment. I have a bit of my own work in progress that I'd like to commit first.

PR #12, opened 2020-01-08
Obsolete blocks

towerofnix:
(PullJosh/scratch-js#42 must be merged before this PR.)

This PR completes support for loading obsolete blocks and adds support for exporting them to scratch-js. Obsolete blocks are defined as the following (mostly copied [from here](https://github.com/LLK/scratch-vm/issues/355#issuecomment-379399646)):

- [x] `doForLoop`: ___. See [here](https://github.com/LLK/scratch-vm/blob/develop/src/blocks/scratch3_control.js#L84-L97) for its most up-to-date behavior (it was kinda glitchy in 2.0 IIRC).
- [x] `doWhile`: ___. [Its behavior](https://github.com/LLK/scratch-flash/blob/9fbac92ef3d09ceca0c0782f8a08deaa79e4df69/src/interpreter/Interpreter.as#L456) is exactly the opposite of `doUntil` (i.e. "repeat until"): while the condition is true (rather than false), run the contained blocks.
- [x] `warpSpeed` ("all at once"): ___. [Its behavior](https://github.com/LLK/scratch-flash/blob/9fbac92ef3d09ceca0c0782f8a08deaa79e4df69/src/interpreter/Interpreter.as#L542-L546) is to run the contained blocks as they are (don't actually go into "warp speed", i.e. no-refresh). ("all at once: { move 10 steps, turn 15 degrees }" is functionally equivalent to "move 10 steps, turn 15 degrees".)
- [x] [Counter blocks](https://github.com/LLK/scratch-flash/blob/9fbac92ef3d09ceca0c0782f8a08deaa79e4df69/src/Specs.as#L378-L381): ___.
  - The counter value is a global value that defaults to zero.
  - `counter`: Return the counter value.
  - `incr counter`: Increment the counter's value by one.
  - `clear counter`: Set the counter value to zero.
- [x] [Scroll blocks](https://github.com/LLK/scratch-flash/blob/9fbac92ef3d09ceca0c0782f8a08deaa79e4df69/src/Specs.as#L386-L391): ___. These are all no-ops.
- [x] ["hide all sprites", "user id"](https://github.com/LLK/scratch-flash/blob/9fbac92ef3d09ceca0c0782f8a08deaa79e4df69/src/Specs.as#L393-L395), and those suspicious "setStretchTo" and "changeStretchBy" blocks: ___. Also no-ops.
- ["loud?"](https://github.com/LLK/scratch-flash/blob/96a2a9a7fca2d042da25dc0d4423900163ab4f33/src/Specs.as#L366): ___. [Its behavior](https://github.com/LLK/scratch-flash/blob/996a2911dd425892f0b4224f5998bce38f964481/src/scratch/ScratchRuntime.as#L1125) is to return whether or not the sound level (i.e. result of "loudness") is greater than 10 (so it's identical to `loudness > 10`). (Not a checkbox since "loudness" hasn't been implemented in scratch-js yet.)
- [Various 1.4 blocks](https://github.com/LLK/scratch-flash/blob/9fbac92ef3d09ceca0c0782f8a08deaa79e4df69/src/Specs.as#L368-L375) that should automatically be converted to their 2.0+ forms. (Not a checkbox since this will be looked at separately from other obsolete blocks - sb-edit doesn't support 1.4 import yet and I'm pretty sure any projects saved in 2.0 or 3.0 will automatically have converted these to the 2.0+ form.)
  - `abs (val)` -> `("abs") of (val)`
  - `sqrt (val)` -> `("sqrt") of (val)`
  - `stop script` -> `stop ("this script")`
  - `stop all` -> `stop ("all")`
  - `switch to background (bg)` -> `switch backdrop to (bg)`
  - `next background` -> `next backdrop`
  - `forever if (cond): (script)` -> `forever: if (cond): (script)`

This PR resolves PullJosh/scratch-js#41.

### DIFF
--- a/src/Block.ts
+++ b/src/Block.ts
@@ -138,6 +138,7 @@ export type KnownBlock =
   | BlockBase<OpCode.control_create_clone_of, { CLONE_OPTION: BlockInput.CloneTarget }>
   | BlockBase<OpCode.control_delete_this_clone, {}>
   | BlockBase<OpCode.control_for_each, { VARIABLE: BlockInput.Variable, VALUE: BlockInput.Number, SUBSTACK: BlockInput.Blocks }>
+  | BlockBase<OpCode.control_while, { CONDITION: BlockInput.Boolean, SUBSTACK: BlockInput.Blocks }>
 
   // Sensing
   | BlockBase<OpCode.sensing_touchingobject, { TOUCHINGOBJECTMENU: BlockInput.TouchingTarget }>
--- a/src/Block.ts
+++ b/src/Block.ts
@@ -138,8 +138,11 @@ export type KnownBlock =
   | BlockBase<OpCode.control_create_clone_of, { CLONE_OPTION: BlockInput.CloneTarget }>
   | BlockBase<OpCode.control_delete_this_clone, {}>
   | BlockBase<OpCode.control_all_at_once, { SUBSTACK: BlockInput.Blocks }>
-  | BlockBase<OpCode.control_for_each, { VARIABLE: BlockInput.Variable, VALUE: BlockInput.Number, SUBSTACK: BlockInput.Blocks }>
-  | BlockBase<OpCode.control_while, { CONDITION: BlockInput.Boolean, SUBSTACK: BlockInput.Blocks }>
+  | BlockBase<
+      OpCode.control_for_each,
+      { VARIABLE: BlockInput.Variable; VALUE: BlockInput.Number; SUBSTACK: BlockInput.Blocks }
+    >
+  | BlockBase<OpCode.control_while, { CONDITION: BlockInput.Boolean; SUBSTACK: BlockInput.Blocks }>
 
   // Sensing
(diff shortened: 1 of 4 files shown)


PullJosh:
Is there a project URL that I can use to test this?

towerofnix:
@PullJosh Oops, I should've linked these in the PR! I've made several projects to test out these blocks, namely: [For-each demo](https://scratch.mit.edu/projects/357632121/), [While demo](https://scratch.mit.edu/projects/357653009/), [Counter demo](https://scratch.mit.edu/projects/357952541/), and [Scroll demo](https://scratch.mit.edu/projects/357968048/).

Each project has a detailed expected-behavior description and a link to implementation code in 3.0, so they should be useful for comparing behavior.

PullJosh:
Looks good for all the projects you've sent. Are there demos of any of the other blocks?

towerofnix:
@PullJosh I didn't bother throwing together demo projects for the other blocks since they're all no-ops, and some I believe don't even load in 3.0 properly but are specified opcodes & 2.0 blocks.

PullJosh:
Thanks for making this happen! 🎉

PR #17, opened 2020-01-12
Support dynamic costume switching

towerofnix:
I.e, support reporters in the switch-costume and switch-backdrop blocks.

This removes the "unique costume name" code which fundamentally prevents dyanmic costume switching, since it renames costumes in ways the project source code generally doesn't expect.

@PullJosh since you wrote that code originally, do you think this is a safe change? I don't understand its original purpose myself, so I just replaced it with this simpler code.

This makes PullJosh/scratch-js#45 have effect—without this PR, scripts such as "switch costume to (any reporter)" just get converted to `this.costume = undefined`—but the two PRs are totally independent of each other, merging-wise.

### DIFF
--- a/src/__tests__/__snapshots__/compilesb3.test.ts.snap
+++ b/src/__tests__/__snapshots__/compilesb3.test.ts.snap
@@ -14,10 +14,10 @@ export default class Abby extends Sprite {
     super(...args);
 
     this.costumes = [
-      new Costume(\\"abbyA\\", \\"./Abby/costumes/abbyA.svg\\", { x: 31, y: 100 }),
-      new Costume(\\"abbyB\\", \\"./Abby/costumes/abbyB.svg\\", { x: 31, y: 100 }),
-      new Costume(\\"abbyC\\", \\"./Abby/costumes/abbyC.svg\\", { x: 32, y: 100 }),
-      new Costume(\\"abbyD\\", \\"./Abby/costumes/abbyD.svg\\", { x: 32, y: 101 })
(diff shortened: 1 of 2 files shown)


PullJosh:
Hmm... The section of the code that generates names is to ensure that all identifiers used in projects (such as sprite names, variable names, custom block names, costume names, etc.) are acceptable javascript identifiers (camel cased). Making the names camel-cased means that we need to reimplement *unique* name constraints because modifying existing names could cause a collision. (Like if a project contained one sprite called `my sprite` and one called `mySprite`.)

**All of that being said, I'm not sure why costume names would need to be valid JS identifiers.** iirc, in the current API costumes are only ever referred to using string ids, which makes this whole process pointless for costumes (and, for that matter, sounds). Am I overlooking some detail that past me considered?

towerofnix:
>All of that being said, I'm not sure why costume names would need to be valid JS identifiers. iirc, in the current API costumes are only ever referred to using string ids, which makes this whole process pointless for costumes (and, for that matter, sounds). Am I overlooking some detail that past me considered?

This is exactly what I was thinking! :P Even in cases where you're referencing a sprite by a static name for some reason (very rare in Scratch, since most sprite dropdowns accept reporters - which means those need to have non-transformed names too, since you can dynamically access them just like costumes), you can just reference them with square bracket syntax - `this.sprites["Scratch Cat"]` for instance.

Yeah, the output code is a little less easy to work with, since it may use non-camelcase names - but that's necessary in order to support dynamic accessing of costumes, sounds, sprites, etc.

PR #21, opened 2020-01-18
scratch-js: support dynamic sounds & sound objects

towerofnix:
As corresponding to #17, this PR removes the code which transformed costume names, since we always refer to sounds by string-style accessing anyway (`getSoundByName`, `startSound`, etc). This enables accessing sounds dynamically, e.g. "play sound (pick random 0 to 4)". It also implements the code for `sound_playuntildone` and `sound_playsound` (the same way that PR does).

This should be merged after PullJosh/scratch-js#49, since it stores sounds on the new `target.sounds` property implemented there.

/cc @PullJosh - btw, do you think we should start organizing PRs and commits with prefixes like "scratch-js: ..." or "scratchblocks: ..."? It would help for reading/searching through commits and PRs, and makes it more obvious what's referring to what.

### DIFF
--- a/src/__tests__/__snapshots__/compilesb3.test.ts.snap
+++ b/src/__tests__/__snapshots__/compilesb3.test.ts.snap
@@ -6,7 +6,8 @@ Object {
   Sprite,
   Trigger,
   Costume,
-  Color
+  Color,
+  Sound
 } from \\"https://pulljosh.github.io/scratch-js/scratch-js/index.mjs\\";
 
 export default class Abby extends Sprite {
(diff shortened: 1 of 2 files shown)


PullJosh:
> do you think we should start organizing PRs and commits with prefixes like "scratch-js: ..." or "scratchblocks: ..."? It would help for reading/searching through commits and PRs, and makes it more obvious what's referring to what.

Maybe. We can also try tags. (One, the other, or both are all options.) Are you able to add tags to your own issues/PRs, or is it just me?

towerofnix:
>Maybe. We can also try tags. (Tags, a prefix, or both are all options.) Are you able to add tags to your own issues/PRs, or is it just me?

Just you! It would be nice if GitHub let owners of a repo mark certain tags as being addable by PR creators; as is, only people with collaboration access are able to do that. (Obviously you don't want to let users add tags like "high prio" to their issues and PRs, but other tags like categorization would be handy. :P)

PullJosh:
> Just you!

Interesting. I think the prefix is a good idea, then.

towerofnix:
>Actually, it looks like the snapshot tests still need to be updated. npm test -- -u

Whoops! Done. This *would* be caught by a CI setup, y'know.... ::::P

PR #22, opened 2020-01-20
License under MIT instead of ISC, and update package.json to match scratch-js

towerofnix:
Same deal as PullJosh/scratch-js#50, though here we're changing the license from ISC to MIT, to keep things consistent across libraries.


PullJosh:
Pretty sure `contributors` should be an array, not an object. :P

towerofnix:
Ohh, um.

I forgot to push 32fa7dcdda4202eb8bdbf60174fb4bb6e5ace821, didn't I.

Oops^2. Thanks! :P

PR #33, opened 2020-02-28
Implement SB3 output

### LINKED ISSUE #28: SB3 Output
Input an SB3, output an SB3! Here's the todo list:

- [x] Loads of research
- [x] Sprite data & costumes
- [x] Most basic blocks
- [x] Structural changes to support sb3 (IDs for sounds & costumes, minor name changes for consistency, etc)
- [x] The rest of the input types & shadow blocks
  - **Data gathering:**
  - [x] Per-block input to default value mapping, to fill in obscured shadow blocks (this isn't sb3-specific)
  - [x] Per-block input to shadow block (primitive & non-primitive) mapping
  - [x] Add remaining extension opcodes (music, etc) to known blocks & above mappings
    - AFAIK extensions always have primitive types & no fields
    - Above might be false for micro:bit grid input though? micro:bit doesn't exist in sb-edit yet anyway, but investigate how that input works to get a better understanding of extensions
  - **Logic:**
  - [x] Fill in primitive values for unobscured shadow blocks
  - [x] Fill in non-primitive values for unobscured shadow blocks
  - [x] Fill in basic default values for obscured shadow blocks
  - [x] For obscured costumes & sounds, replace default value with the corresponding first name in the sprite
  - [x] Construct broadcast ID-name mapping when broadcast inputs encountered (they're internally stored as variables now)
- [x] Custom procedures
- [x] Code finalization
  - [x] Testing lots and lots of projects
  - [x] Extend code documentation
  - [x] Add snapshot test
  - [x] Prettify code
  - [x] Make pull request

There is already a partially-complete implementation for sb3 output; it should be completed, so that projects loaded from any format can be converted into the most common output format. (As of present, our only ([mostly](https://github.com/PullJosh/scratch-js/issues/18)) complete output is scratch-js.) Implementing a working SB3 output will also let us have a basis for implementing new similar output formats, like SB2: the code structure will likely be similar, and the understanding from solving problems in SB3 will probably be useful going forward.

I've begun an SB2 output format [here](https://github.com/towerofnix/sb-edit/tree/sb2), but I ran into some peculiarities whose solutions would probably be highlighted by completing an SB3 format. SB3 is also the format which sb-edit is natively based on, with its selection of opcodes and code structure—while implementing SB3 prompts questions like "how do we handle assets?", SB2 implies those in addition to complex questions about backporting blocks, etc. It'd be better to handle one bunch of questions at a time, so: SB3 before SB2.

I'm happy to start work on this myself, since I've already worked with sb-edit a fair bit and don't have any current project in relation to it.

towerofnix:
Resolves #28.

This is a fairly large pull request.

It would be basically impossible to capture everything that's been implemented here in any efficient manner; but, in essence, this PR does two things:

1. Change interfaces & structures to closer match SB3 (which sb-edit is based on), and add some convenient APIs for usage in the serialization code.
2. Implement the SB3 output format.

About 3000 lines of the diff are dedicated exclusively to the new snapshot test; another thousand is for newly added or updated block data mappings. There are a few hundred lines of documentation in `io/sb3/toSb3.ts`, which should be used as the base for understanding the changes in this PR.
(#32 has some handy visuals on how inputs work too.)

No pressure to get this reviewed quickly! It's big. I did end up [taking some artistic liberties](https://github.com/PullJosh/sb-edit/issues/28#issuecomment-584848945); hopefully that's alright. :P None of the outward-facing APIs have changed (e.g. `toScratchJS`), and only the sb3 output has had its behavior updated, so the library should continue to work in all existing use cases without requiring any changes.


src/Costume.ts
```diff
@@ -1,22 +1,40 @@
+import { generateId } from "./util/id";
+
+type CostumeDataFormat = "png" | "svg" | "jpeg" | "jpg" | "bmp" | "gif";
```
adroitwhiz: I've never seen `.bmp` costumes before! Does 2.0 allow them?

src/List.ts
```diff
@@ -1,8 +1,10 @@
-import { generateID } from "./util/id";
+import { generateId } from "./util/id";
+
+type ScalarValue = string | number | boolean;
```
adroitwhiz: Is this type duplicated between `List.ts` and `Variable.ts`?

src/io/sb3/fromSb3.ts
```diff
@@ -13,55 +13,60 @@ import { Sprite, Stage } from "../../Target";
 import Variable from "../../Variable";
 import Script from "../../Script";
 
-type getAsset = (info: {
+type getAssetData = (info: {
```
adroitwhiz: An explicit `Asset` type may be helpful

src/io/sb3/fromSb3.ts
```diff
@@ -85,93 +90,104 @@ function getBlockScript(blocks: { [key: string]: sb3.Block }) {
...
-              addInput(inputName, {
-                type: "number",
-                value: isNaN(parseFloat(value[1])) ? (value[1] as string) : parseFloat(value[1])
-              });
+            case BIS.MATH_NUM_PRIMITIVE:
+            case BIS.POSITIVE_NUM_PRIMITIVE:
+            case BIS.WHOLE_NUM_PRIMITIVE:
+            case BIS.INTEGER_NUM_PRIMITIVE: {
+              let storedValue: string | number = value[1];
+              if (!isNaN(parseFloat(storedValue as string))) {
```
adroitwhiz: `parseFloat` [will parse things like `"1x"` as `1`](https://stackoverflow.com/questions/12227594/which-is-better-numberx-or-parsefloatx). Not sure if that's what Scratch does in this situation (or if that path will ever be hit), but it may be better to use a `Number` cast.

src/io/sb3/interfaces.ts
```diff
@@ -0,0 +1,502 @@
...
+
+  shadow: boolean;
+  topLevel: boolean;
+
+  x?: number;
+  y?: number;
+
+  mutation?: {
+    tagName: string;
+    children: any[]; // TODO: What is this?
```
adroitwhiz: I think `mutation` is a JSON version of an XML node, and so `children` is an array of other XML nodes.

src/Variable.ts
```diff
@@ -13,31 +17,25 @@ export default class Variable {
...
-    },
-    id?: string
-  ) {
+    visible?: boolean;
+    mode?: "default" | "slider" | "large";
+    x?: number;
+    y?: number;
+    sliderMin?: number;
+    sliderMax?: number;
+    isDiscrete?: boolean;
```
adroitwhiz: Will variables (and lists, for that matter) ever be initialized without these values provided?

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1124 @@
...
+
+    // Just for debugging!
+    const entryKeys = Object.keys(inputEntries);
+    const fieldEntryKeys = Object.keys(sb3.fieldTypeMap[block.opcode] || {});
+    const missingEntries = Object.keys(inputs).filter(
+      inp => !(entryKeys.includes(inp) || fieldEntryKeys.includes(inp))
+    );
+    for (const key of missingEntries) {
+      warn(`Missing entry for input ${key} on ${block.opcode} (${block.id} in ${target.name})`);
+    }
```
adroitwhiz: How expensive is this? Would it be beneficial to only do all this work if `warn` is passed?

src/Costume.ts
towerofnix: Oh, interesting! This line is taken [from scratch-parser](https://github.com/LLK/scratch-parser/blob/cee1200c858853c0f498d3d01dfbd02f792c6717/lib/sb3_definitions.json#L49-L52). I haven't actually tested bmp costumes in the editor; I'm just matching the official validation format here.

src/List.ts
towerofnix: Yep - as of 7db56d4, the two modules are now combined, since we export multiple things from files like `Target` or `Block` already.

src/io/sb3/interfaces.ts
towerofnix: I want to make the mutation attribute type better in any case - allowing for the specific mutations which Scratch supports. Right now it's a bit of a mish-mash of properties that don't matter as much as they should, typing-wise.

src/Variable.ts
towerofnix: Probably not, but otherwise the default values listed above mean nothing.
/cc @PullJosh for thoughts here!

src/io/sb3/toSb3.ts
towerofnix: Oh, good point! Definitely.

Buuuuuuuut this particular debugging code should probably be removed, because it's only necessary for working out new blocks - AFAICT I've tested all the blocks supported by sb-edit already, so it's not needed anymore.

src/io/sb3/interfaces.ts
towerofnix: I couldn't figure out how to make a clean type format for this, so I just boiled it to `{[attribute: string]: any}`, and have documented what it represents.

src/io/sb3/fromSb3.ts
```diff
@@ -198,7 +213,7 @@ function getBlockScript(blocks: { [key: string]: sb3.Block }) {
                 }) as Block
```
adroitwhiz: Is the `// TODO: Set "parent"` above this line something that should be taken care of?

src/io/sb3/fromSb3.ts
```diff
@@ -1,67 +1,71 @@
...
             md5: costumeData.assetId,
             ext: costumeData.dataFormat,
             spriteName: target.name
-          })
+          }),
+
+          md5: costumeData.assetId,
+          ext: costumeData.dataFormat,
+
+          bitmapResolution: costumeData.bitmapResolution || 1,
```
adroitwhiz: Scratch 3.0 treats an unspecified `bitmapResolution` as 2.

src/io/sb3/fromSb3.ts
```diff
@@ -1,67 +1,71 @@
...
+          centerX: costumeData.rotationCenterX || 0,
+          centerY: costumeData.rotationCenterY || 0
```
adroitwhiz: I believe 3.0 automatically calculates rotation centers for costumes which don't have them, but that could get messy very quickly (e.g. calculating SVG rotation centers requires calculating the costume's bounds). Maybe add a note for now that this doesn't match 3.0's behavior.

src/io/sb3/fromSb3.ts
adroitwhiz: [Here's the code for reference](https://github.com/LLK/scratch-vm/blob/01cb244e3dc5296183ad67a6ee7cd80c4ca1b355/src/import/load-costume.js#L133). `scale` refers to the scaling factor, so costumes will only be scaled up to double resolution if they are explicitly stated to have a resolution of 1.

src/io/sb3/fromSb3.ts
adroitwhiz: Or just allow `centerX` and `centerY` to be nullable and pass the complexity onto the Scratch implementation that uses `sb-edit`

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
...
+    return { blockData, blockId };
+  }
+
+  interface CustomBlockData {
+    args: Array<{
+      default: string;
+      id: string;
+      name: string;
+      type: "boolean" | "numberOrString";
+    }>;
```
adroitwhiz: Could
```ts
    {
      default: string;
      id: string;
      name: string;
      type: "boolean" | "numberOrString";
    }
```

be made into a `CustomBlockArg` interface so that `args` below can be an `Array<CustomBlockArg>`?

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
...
+
+      currentCostume: target.costumeNumber,
+      layerOrder: target.layerOrder,
+      volume: target.volume,
+
+      blocks,
+      broadcasts,
+
+      // sb-edit doesn't support comments (as of feb 12, 2020)
+      comments: {},
```
adroitwhiz: nit: add a `TODO` so it's easier to search for? (plus eslint can optionally warn on `TODO`s)

src/Sound.ts
```diff
@@ -1,22 +1,38 @@
...
   public sampleCount: number;
   public sampleRate: number;
-  public format: string;
-  public asset: any;
 
   constructor(options: {
     name: string;
+    id?: string;
+
+    dataFormat: SoundDataFormat;
```
adroitwhiz: What's the difference between `dataFormat` and `ext`?

src/__tests__/compilesb3.test.ts
```diff
@@ -3,16 +3,22 @@ import { Project } from "..";
 import * as fs from "fs";
 import * as path from "path";
 
-test("sb3 -> scratch-js", async () => {
-  const file = fs.readFileSync(path.join(__dirname, "test.sb3"));
-  const project = await Project.fromSb3(file);
+async function loadProject(filename) {
+  const file = fs.readFileSync(path.join(__dirname, filename));
+  return await Project.fromSb3(file);
```
adroitwhiz: `return await` is [apparently unnecessary/an antipattern](https://eslint.org/docs/rules/no-return-await).

src/io/sb3/fromSb3.ts
```diff
@@ -482,8 +427,8 @@ export async function fromSb3JSON(json: sb3.ProjectJSON, options: { getAsset: ge
 export default async function fromSb3(fileData: Parameters<typeof JSZip.loadAsync>[0]): Promise<Project> {
   const inZip = await JSZip.loadAsync(fileData);
   const json = await inZip.file("project.json").async("text");
-  const getAsset = async ({ md5, ext }) => {
+  const getAssetData = async ({ md5, ext }) => {
     return await inZip.file(`${md5}.${ext}`).async("arraybuffer");
```
adroitwhiz: same here w.r.t `return await`

src/io/sb3/fromSb3.ts
```diff
@@ -482,8 +427,8 @@ export async function fromSb3JSON(json: sb3.ProjectJSON, options: { getAsset: ge
...
   };
-  return await fromSb3JSON(JSON.parse(json), { getAsset });
+  return await fromSb3JSON(JSON.parse(json), { getAssetData });
```
adroitwhiz: same here w.r.t `return await`

src/io/sb3/fromSb3.ts
```diff
@@ -416,11 +360,12 @@ export async function fromSb3JSON(json: sb3.ProjectJSON, options: { getAsset: ge
...
     stage: new Stage({
       name: stage.name,
-      costumes: await extractCostumes(stage, options.getAsset),
+      costumes: await extractCostumes(stage, options.getAssetData),
       costumeNumber: stage.currentCostume,
-      sounds: await extractSounds(stage, options.getAsset),
+      sounds: await extractSounds(stage, options.getAssetData),
       scripts: Object.entries(stage.blocks)
         .filter(([id, block]) => block.topLevel)
+        .filter(([id, block]) => !block.shadow)
```
adroitwhiz:
```diff
-        .filter(([id, block]) => block.topLevel)
-        .filter(([id, block]) => !block.shadow)
+        .filter(([id, block]) => block.topLevel && !block.shadow)
```
Half the filtering, twice\* the performance!


\* 60% of the time, it increases performance every time

src/io/sb3/fromSb3.ts
```diff
@@ -441,9 +386,9 @@ export async function fromSb3JSON(json: sb3.ProjectJSON, options: { getAsset: ge
           async (spriteData: sb3.Sprite) =>
             new Sprite({
               name: spriteData.name,
-              costumes: await extractCostumes(spriteData, options.getAsset),
+              costumes: await extractCostumes(spriteData, options.getAssetData),
               costumeNumber: spriteData.currentCostume,
-              sounds: await extractSounds(spriteData, options.getAsset),
+              sounds: await extractSounds(spriteData, options.getAssetData),
```
adroitwhiz: Is there any benefit to running `extractSounds` and `extractCostumes` in parallel, a la:
```js
const [costumes, sounds] = await Promise.all([
    extractCostumes(spriteData, options.getAssetData),
    extractSounds(spriteData, options.getAssetData)
]);
```

src/io/sb3/fromSb3.ts
```diff
@@ -441,9 +386,9 @@ export async function fromSb3JSON(json: sb3.ProjectJSON, options: { getAsset: ge
...
               scripts: Object.entries(spriteData.blocks)
                 .filter(([id, block]) => block.topLevel)
                 .filter(([id, block]) => !block.shadow)
```
adroitwhiz: same thing w.r.t "half the filtering, twice the performance"

adroitwhiz: in fact, is there any way to de-duplicate this code with the above (`Object.entries(stageData.blocks)` and so forth)?

src/io/sb3/fromSb3.ts
```diff
@@ -85,93 +89,104 @@ function getBlockScript(blocks: { [key: string]: sb3.Block }) {
...
+                ...translateFields(blocks[value].fields, blocks[value].opcode)
+              };
+            }
+          } else {
+            let isBlocks;
+            if (BlockBase.isKnownBlock(block.opcode)) {
+              const { type } = BlockBase.getDefaultInput(block.opcode, inputName) || {};
+              if (type === "blocks") {
+                isBlocks = true;
               }
```
adroitwhiz:
```suggestion
              const defaultInput = BlockBase.getDefaultInput(block.opcode, inputName);
              if (defaultInput && defaultInput.type === "blocks") {
                isBlocks = true;
              }
```
I think this reads a bit clearer

src/io/sb3/fromSb3.ts
```diff
@@ -85,93 +89,104 @@ function getBlockScript(blocks: { [key: string]: sb3.Block }) {
...
+            }
+            isBlocks = isBlocks || inputScript.length > 1;
+            if (isBlocks) {
+              addInput(inputName, { type: "blocks", value: inputScript });
             } else {
               addInput(inputName, { type: "block", value: inputScript[0] });
             }
```
adroitwhiz: If I understand correctly, these cases will only be hit when there's an unknown block. Will that ever actually happen?

src/io/sb3/fromSb3.ts
```diff
@@ -336,7 +284,7 @@ function getBlockScript(blocks: { [key: string]: sb3.Block }) {
   return blockWithNext;
```
adroitwhiz: This may be outside the scope of this PR (if so, I'll create an issue to track it separately).

It looks like the `blockWithNext` function is recursive, and creates a new `Array` object *every time* it recurses. This sounds like extremely bad news for performance, especially with long scripts. Could it be changed to be less "clever" and use an iterative approach?

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
...
+    broadcastIdToName = Object.assign(
+      {},
+      ...Object.entries(broadcastIdToName)
+        .sort(([, name1], [, name2]) => (name1 < name2 ? -1 : 1))
+        .map(([id, name]) => ({ [id]: name }))
+    );
+
+    // Set the broadcast name used in obscured broadcast inputs to the first
+    // sorted-alphabetically broadcast's name.
+    getBroadcastId.initialBroadcastName = Object.keys(broadcastNameToId)[0] || "message1";
```
adroitwhiz: Object key order is, in some cases, [not guaranteed by the ECMAScript spec](https://stackoverflow.com/questions/30076219/does-es6-introduce-a-well-defined-order-of-enumeration-for-object-properties), so it may be better to use some other data structure if you want to guarantee a specific sort order.

In particular, it looks like here, you're only sorting these objects so that you can set `initialBroadcastName` to the one that comes alphabetically "first", and so you could easily replace the sorting altogether with a "find alphabetically lowest" method.

adroitwhiz: Also, modifying a function's prototype to add a custom property seems kinda janky

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
...
+      for (const block of target.blocks) {
+        if (block.opcode === OpCode.event_whenbroadcastreceived) {
+          if (block.inputs.BROADCAST_OPTION.type === "broadcast") {
+            getBroadcastId(block.inputs.BROADCAST_OPTION.value);
+          }
+        } else if (block.opcode === OpCode.event_broadcast || block.opcode === OpCode.event_broadcastandwait) {
+          if (block.inputs.BROADCAST_INPUT.type === "broadcast") {
+            getBroadcastId(block.inputs.BROADCAST_INPUT.value);
+          }
+        }
```
adroitwhiz:
```suggestion
        if (
          block.opcode === OpCode.event_whenbroadcastreceived ||
          block.opcode === OpCode.event_broadcast ||
          block.opcode === OpCode.event_broadcastandwait
        ) {
          const broadcastInput = (block.opcode === OpCode.event_whenbroadcastreceived) ?
            block.inputs.BROADCAST_OPTION :
            block.inputs.BROADCAST_INPUT;

          if (broadcastInput.type === "broadcast") {
            getBroadcastId(broadcastInput.value);
          }
        }
```

This looks more DRY to me but it's a matter of preference.

adroitwhiz: Also, under what conditions is the `type` not `"broadcast"`?

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
...
+  }
+
+  function collectCustomBlockData(target: Target): GetCustomBlockData {
+    // Parse the scripts in a target, collecting metadata about each custom
+    // block's arguments and other info, and return a function for accessing
+    // the data associated with a particular custom block's proccode.
+    //
+    // It's necesary to collect this data prior to serializing any associated
+    // procedure_call blocks, because they require access to data only found
+    // on the associated procedure_definition. (Specifically, the types of
```
adroitwhiz:
```suggestion
    // procedures_call blocks, because they require access to data only found
    // on the associated procedures_definition. (Specifically, the types of
```

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
...
+
+    const blockData = newBlockData();
+
+    const { getBroadcastId, getCustomBlockData, parent, siblingBlocks, stage, target } = options;
+
+    let nextBlock: Block;
+    let nextBlockId: sb3.Block["next"];
+    if (siblingBlocks) {
+      const thisIndex = siblingBlocks.indexOf(block) + 1;
+      nextBlock = siblingBlocks.find((x, index) => index === thisIndex);
```
adroitwhiz:
```suggestion
      nextBlock = siblingBlocks[thisIndex];
```

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
...
 
-import List from "../../List";
-import Variable from "../../Variable";
+export default function toSb3(options: Partial<ToSb3Options> = {}): ToSb3Output {
+  // Serialize a project. Returns an object containing the text to be stored
+  // in the caller's project.json output file.
 
-export default async function toSb3(project: Project): Promise<ArrayBuffer> {
-  const outZip = new JSZip();
+  const project: Project = this;
```
adroitwhiz: Aliasing `project = this` at the very top of the function and then only using `project` at the very bottom decreases readability IMO.

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
...
+      const input = inputs[key];
+      // Fields are stored as a plain [value, id?] pair.
+      let id: string;
+      switch (input.type) {
+        case "variable":
+          id = getVariableId(input.value, target, stage);
+        case "list":
+          id = getListId(input.value, target, stage);
+        default:
+          id = null;
```
adroitwhiz:
```suggestion
        case "variable":
          id = getVariableId(input.value, target, stage);
          break;
        case "list":
          id = getListId(input.value, target, stage);
          break;
        default:
          id = null;
          break;
```
Watch out for fallthrough!

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
...
-      sampleCount: sound.sampleCount,
-      md5ext: `${sound.md5}.${sound.ext}`
+    if (!fieldEntries) {
+      return result;
+    }
+
+    for (const key of Object.keys(fieldEntries)) {
+      const input = inputs[key];
+      // Fields are stored as a plain [value, id?] pair.
+      let id: string;
```
adroitwhiz: Should this be `string | null`?

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
...
+
+  function applyBlockData(a: BlockData, b?: BlockData) {
+    if (b) {
+      Object.assign(a, b);
     }
   }
 
-  outZip.file("project.json", JSON.stringify(getProjectJSON(project)));
+  function newBlockData(): BlockData {
+    return {};
```
adroitwhiz: What purpose does this function serve?

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
...
+            if (target.costumes[0]) {
+              valueForShadow = target.costumes[0].name;
+            }
+          } else if (entry === OpCode.sound_sounds_menu) {
+            if (target.sounds[0]) {
+              valueForShadow = target.sounds[0].name;
+            }
+          } else if (entry === OpCode.event_broadcast_menu) {
+            valueForShadow = getBroadcastId.initialBroadcastName;
+          }
```
adroitwhiz:
```suggestion
          // Special-case some input opcodes for more realistic initial values.
          switch (entry) {
            case OpCode.looks_costume:
              valueForShadow = target.costumes[0].name;
              break;
            case OpCode.sound_sounds_menu:
              valueForShadow = target.sounds[0].name;
              break;
            case OpCode.event_broadcast_menu:
              valueForShadow = getBroadcastId.initialBroadcastName;
              break;
            default:
              valueForShadow = initialValues[key];
              break;
          }
```

adroitwhiz: Also, what happens to `valueForShadow` if the target has no sounds (or no `initialBroadcastName`)? Will it throw an error since it's trying to access `name`?

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
...
+
+          break;
+        }
+
+        default: {
+          const inputEntries = prop(sb3.inputPrimitiveOrShadowMap, block.opcode);
+
+          const initialValues = {};
+          for (const key of Object.keys(inputEntries)) {
+            initialValues[key] = (BlockBase.getDefaultInput(block.opcode, key) || {}).initial;
```
adroitwhiz:
```suggestion
            const defaultInput = BlockBase.getDefaultInput(block.opcode, key);
            if (defaultInput)) {
              initialValues[key] = defaultInput.initial;
            }
```

Not sure if this changes the behavior since setting object values to `undefined` is not the same as never adding them in the first place, but `JSON.stringify` doesn't serialize `undefined` properties anyway.

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
...
+            );
+            initialValues[arg.id] = prop(
+              {
+                boolean: false,
+                numberOrString: ""
+              },
+              arg.type
+            );
+            constructedInputs[arg.id] = block.inputs.INPUTS.value[i++];
+          }
```
adroitwhiz:
```suggestion
          for (let i = 0; i < args.length; i++) {
            const { type, id } = args[i];
            inputEntries[id] = prop(
              {
                boolean: sb3.BooleanOrSubstackInputStatus,
                numberOrString: BIS.TEXT_PRIMITIVE
              },
              type
            );
            initialValues[id] = prop(
              {
                boolean: false,
                numberOrString: ""
              },
              type
            );
            constructedInputs[id] = block.inputs.INPUTS.value[i];
          }
```

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
...
-      md5ext: `${costume.md5}.${costume.ext}`,
-      dataFormat: costume.ext,
-      rotationCenterX: costume.centerX / 2,
-      rotationCenterY: costume.centerY / 2
+    fieldEntries;
+  }
+
+  function serializeInputsToFields(
+    inputs: { [key: string]: BlockInput.Any },
+    options: SerializeInputsToFields
```
adroitwhiz:
```suggestion
    options: SerializeInputsToFieldsOptions
```

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
...
+  function getListId(listName: string, target: Target, stage: Stage): string | null {
+    // Get the ID associated with the provided list name on either the provided
+    // target or the stage, or null if no matching list is found.
+
+    let list = target.getList(listName);
+    list = list || stage.getList(listName);
+    return list ? list.id : null;
+  }
+
+  interface SerializeInputsToFields {
```
adroitwhiz:
```suggestion
  interface SerializeInputsToFieldsOptions {
```

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
...
-      (lists: sb3.Stage["lists"], list) => ({
-        ...lists,
-        [list.id]: [list.name, list.value]
+
+      data[proccode] = { args, warp };
+    }
+
+    return (proccode: string): CustomBlockData => {
+      return data[proccode];
+    };
```
adroitwhiz: Not entirely sure I agree with the design pattern of dynamically creating getter functions.

Yes, the data structure is now "opaque", but IMO it makes the code much harder to reason about. Instead of taking one *specific* type of block data, `serializeBlock` could theoretically take *any* function that returns `CustomBlockData` in any number of ways. You have to trawl through the code just to figure out that there's only one concrete implementation called in one place.

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
...
+    const mapToIdObject = (
+      values: Array<{ id: string; [propName: string]: any }>,
+      fn: (x: any) => any
+    ): { [key: string]: any } => {
+      const ret = {};
+      for (const object of values) {
+        ret[object.id] = fn(object);
+      }
+      return ret;
+    };
```
adroitwhiz: This is a pretty dense function that could benefit from some documentation.

src/io/sb3/toSb3.ts
adroitwhiz: If this is about functions not needing to know `BlockData`'s internal data structure, why?

`BlockData` matches the `blocks` section of a `.sb3` Target, so unless the Scratch Team decides to make `blocks` an `Array` in all future `.sb3`s, this abstraction seems unnecessary to me: the "internal" data structure isn't actually so internal--it's an essential property of the `.sb3` format itself.

adroitwhiz: You can still capture the type/semantics with
```ts
const blockData: BlockData = {};
```

src/io/sb3/toSb3.ts
adroitwhiz: maybe also rename `thisIndex` to `nextIndex`

adroitwhiz:
Just remembered-- you should probably update the README now that sb3 input + output are both fully supported

src/Sound.ts
towerofnix: I don't think so; AFAIK, there's no difference. Worth removing from Costume too. Thanks!

towerofnix: (I guess my pending comment here disappeared...?) We don't need both - I got rid of ext in Costume too. Thanks!

src/io/sb3/fromSb3.ts
towerofnix: Probably! Updated, thanks. (Commits to be pushed soon™.)

src/io/sb3/fromSb3.ts
towerofnix: Absolutely! I made a function to share all the common properties between the stage and sprites.

src/io/sb3/fromSb3.ts
towerofnix: It's possible that we parse unknown blocks (i.e. `!isKnownBlock(block.opcode)`), in the case of a project containing blocks that sb-edit doesn't specifically recognize yet. These cases could be hit if an extension whose blocks are unknown contained a C block. (I believe such can't happen in Scratch yet, because extensions don't support C blocks, but that will probably change [in the future](https://github.com/LLK/scratch-vm/issues/2120).)

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1114 @@
-const doesThisCodeHaveErrors = "not if I comment it out";
-export default doesThisCodeHaveErrors;
+import Block, { BlockBase, KnownBlock, ProcedureBlock } from "../../Block";
+import Costume from "../../Costume";
+import Project, { TextToSpeechLanguage } from "../../Project";
+import Script from "../../Script";
+import Sound from "../../Sound";
```
PullJosh: It seems like `KnownBlock`, `ProcedureBlock`, `Costume`, `Script`, and `Sound` are all unused imports.
```suggestion
import Block, { BlockBase } from "../../Block";
import Project, { TextToSpeechLanguage } from "../../Project";
```

src/io/sb3/toSb3.ts
towerofnix: 0_0 thanks!

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1120 @@
...
+
+          shadow: true,
+          topLevel: false
+        };
+
+        shadowValue = id;
+      }
+    }
+
+    return { shadowValue };
```
adroitwhiz: Now that this only returns one value, you don't need to structure/destructure it

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1120 @@
...
+              },
+              type
+            );
+            initialValues[id] = prop(
+              {
+                boolean: false,
+                numberOrString: ""
+              },
+              type
+            );
```
adroitwhiz:
```suggestion
            switch (type) {
              case "boolean":
                inputEntries[id] = sb3.BooleanOrSubstackInputStatus;
                initialValues[id] = false;
                break;
              case "numberOrString":
                inputEntries[id] = BIS.TEXT_PRIMITIVE;
                initialValues[id] = "";
                break;
            }
```

I *think* this does the same thing. Not sure if `args[i].type` is guaranteed to be `"boolean" | "numberOrString"`; if it's not then add a `default` which throws.

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1120 @@
...
+      [key in keyof PassedInputs]: any;
     };
-    return sb3Costume;
-  });
-}
+  }
+
+  function serializeInputsToInputs<PassedInputs extends { [key: string]: BlockInput.Any }>(
+    inputs: PassedInputs,
+    options: SerializeInputsToInputsOptions<PassedInputs>
```
adroitwhiz: Is there any info in `inputEntries` that doesn't exist in `inputs`?

src/io/sb3/toSb3.ts
adroitwhiz: Also, here the initial value is `false` (a boolean) but [here](https://github.com/PullJosh/sb-edit/pull/33/files#diff-9d556895c6d9cbaaf5813685827b42acR849) it's `"false"` (a string). Is this intentional?

src/io/sb3/toSb3.ts
towerofnix: Yep - this reminds me `inputEntries` doesn't currently have a type definition. I should definitely give it one. Essentially, it refers to a value in `inputPrimitiveOrShadowMap` (see interfaces.ts). It describes how the input shadow should be serialized (i.e. as which kind of primitive or which menu opcode), which is data that sb-edit doesn't provide elsewhere.

src/io/sb3/toSb3.ts
towerofnix: re: second comment - That's a good point. The initial value for a boolean...actually doesn't matter, because when empty boolean inputs are serialized, the passed value isn't considered at all. As such, it's technically not necessary to provide an initial value here at all. I haven't tested (I'll get back to messing with code tomorrow), but simply not setting `initialValues[id]` should be valid code.

As for whether it's necessary in argumentids (which is serialized directly from the `CustomBlockArg` data) - I'm not sure. I provide `"false"` because that's the value that Scratch 3.0 outputs. I would suggest it might be possible to simply output undefined, ala `["",,""]` in custom blocks, but JSON doesn't support output like that, so it's basically a moot point. Better to match the output of 3.0 in the same situation.

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1120 @@
...
+              target
+            });
+          }
+        }
+
+        if (obscuringBlockValue) {
+          resultInputs[key] = [BIS.INPUT_DIFF_BLOCK_SHADOW, obscuringBlockValue, shadowValue];
+        } else {
+          resultInputs[key] = [BIS.INPUT_SAME_BLOCK_SHADOW, shadowValue];
+        }
```
adroitwhiz:
```suggestion
        if (input.type === "block") {
          let obscuringBlockValue;

          if (input.value.opcode === OpCode.data_variable) {
            const variableName = input.value.inputs.VARIABLE.value;
            const variableId = getVariableId(variableName, target, stage);
            obscuringBlockValue = [BIS.VAR_PRIMITIVE, variableName, variableId];
          } else if (input.value.opcode === OpCode.data_listcontents) {
            const listName = input.value.inputs.LIST.value;
            const listId = getListId(listName, target, stage);
            obscuringBlockValue = [BIS.LIST_PRIMITIVE, listName, listId];
          } else {
            obscuringBlockValue = serializeBlock(input.value, {
              blockData,
              getBroadcastId,
              customBlockDataMap,
              parent: block,
              stage,
              target
            });
          }

          resultInputs[key] = [BIS.INPUT_DIFF_BLOCK_SHADOW, obscuringBlockValue, shadowValue];
        } else {
          resultInputs[key] = [BIS.INPUT_SAME_BLOCK_SHADOW, shadowValue];
        }
```

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1120 @@
...
+    //
+    // Blocks may also have a "mutation" field. This is an XML attribute
+    // mapping containing data specific to a particular instance of a block
+    // that wouldn't fit on the block's input and field mappings. Specific
+    // details may vary greatly based on the opcode.
+
+    const { blockData, stage, target, getBroadcastId, customBlockDataMap } = options;
+
+    const { fields } = serializeInputsToFields(block.inputs, {
+      fieldEntries: sb3.fieldTypeMap[block.opcode],
```
adroitwhiz: Why pass `fieldEntries` here instead of just accessing `sb3.fieldTypeMap[block.opcode]` in `serializeInputsToFields`?

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1120 @@
...
 
-import List from "../../List";
-import Variable from "../../Variable";
+    const { fieldEntries, stage, target } = options;
 
-export default async function toSb3(project: Project): Promise<ArrayBuffer> {
-  const outZip = new JSZip();
+    const result = {
+      fields: {}
+    };
```
adroitwhiz: Same here re: structuring/destructuring roundtrip when only one value is returned

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1120 @@
...
+          break;
+        case "list":
+          id = getListId(input.value, target, stage);
+          break;
+        default:
+          id = null;
+          break;
+      }
+      result.fields[key] = [input.value, id];
+      continue;
```
adroitwhiz: unnecessary `continue`?

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1120 @@
...
+            );
+            return null;
+          }
+
+          const { args, warp } = customBlockData;
+
+          mutation = {
+            tagName: "mutation",
+            children: [],
+            proccode: block.inputs.PROCCODE.value,
```
adroitwhiz:
```suggestion
            proccode,
```

src/io/sb3/toSb3.ts
towerofnix: Nice, thanks! Would a switch/case be better here?

adroitwhiz: Probably

src/io/sb3/toSb3.ts
towerofnix: There's no reason we're required to pass it here, since this is the only call to `serializeInputsToFields` in toSb3; I kept it for consistency, because the corresponding function for inputs requires `inputEntries` be passed (it's called below with dynamically constructed contents for custom blocks).

adroitwhiz: IMO it's a bit misleading/confusing to pass it in since it implies a "dynamic-ness" that doesn't exist. Inputs != fields anyways, so I think it's fine if the signatures are different between the two functions.

src/io/sb3/toSb3.ts
towerofnix: Apparently! It's been there since the first sb3 commit, so I have no idea how it ended up there. I'll get it removed :P

src/io/sb3/toSb3.ts
towerofnix: Sure, sounds good to me. TY for the commits btw!

adroitwhiz:
Also just a heads-up @towerofnix -- I force-pushed to this branch so make sure to `git reset --hard` to this branch (the "origin" one on your end, I think) before doing further work.

src/io/sb3/toSb3.ts
adroitwhiz: On second thought, it's better to pass in `fieldEntries` than to pass in the entire `block` so that `serializeInputsToFields` can access `block.opcode`.

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1112 @@
...
+        broadcastNameToId[name] = id;
+        broadcastIdToName[id] = name;
+      }
+      return broadcastNameToId[name];
     };
-    return sb3Sprite;
-  });
+
+    // Preemptively parse through all the broadcast inputs in the project and
+    // generate a broadcast ID for each of them.
```
adroitwhiz: What's the reasoning behind doing this preemptively when it looks like `getBroadcastId` handles this correctly if doing this non-preemptively?

towerofnix: We preemptively do so that we can find the alphabetically first broadcast name below - but you're right that getBroadcastId will handle it irregardless of if we parse preemptively, so I clarified the logic in 5144e3918d05e9832f4357f45bf23d5095924392.

towerofnix:
Alright, a few commits later and I think this is ready for merge. I made some minor improvements to code logic all over - see above commits - and [made sb3 output predictable](https://github.com/PullJosh/sb-edit/pull/33/commits/8bfe71ad716310a6999b52f26016efa20c483dd9), allowing the snapshot test (now [updated](https://github.com/PullJosh/sb-edit/pull/33/commits/f67ae3e74820fd6b79df39358512e22969b8621c)) to be reliable.

src/io/sb3/toSb3.ts
```diff
@@ -1,163 +1,1094 @@
...
+    // stored inline with the input they correspond to, not as separate blocks
+    // with IDs.
+    //
+    // Within this code, we use the primimtiveOrOpCode option to determine how
+    // the shadow should be serialized. If it is a number, it's referring to a
+    // "primitive", the term uses for shadows when they are in the compressed
+    // form. If it's a string, it is a (shadow) block opcode, and should be
+    // serialized in the expanded form.
+
+    const { blockData, initialBroadcastName, parentId, primitiveOrOpCode, shadowId } = options;
```
adroitwhiz: ESLint says `initialBroadcastName` is unused here

src/io/sb3/interfaces.ts
```diff
@@ -0,0 +1,497 @@
+import { OpCode } from "../../OpCode";
+import { KnownBlock, ProcedureBlock } from "../../Block";
+import { TextToSpeechLanguage } from "../../Project";
+import * as BlockInput from "../../BlockInput";
```
adroitwhiz: Can you import this as `_BlockInput` instead? You're defining your own `BlockInput` type and exporting it in this same file, and under some circumstances, the TypeScript compiler will error out if you import something and export something else under the same name.

towerofnix:
@adroitwhiz Fixed those and pushed! Thanks.

adroitwhiz:
Merged!

PR #41, opened 2020-03-04
scratch-js: don't restringify number inputs

### LINKED ISSUE #9: Recognize hex notation (0xff0000) as a number when exporting scratch-js
See https://github.com/PullJosh/scratch-js/commit/a2ff1a55f5f8d8b57bd35ece4ff77bdc8437650b#commitcomment-36606861

image
In the situation above, the compiled scratch-js code should be
```js
this.vars.myVariable = 0x7f7f7f; // good
```
rather than
```js
this.vars.myVariable = "0x7f7f7f"; // bad
```

adroitwhiz:
Resolves #9 

Currently, `sb-edit` compiles this:
image
into this:
```js
this.stage.vars.myVariable = 8355711;
```

With this PR, it compiles to this:
```js
this.stage.vars.myVariable = 0x7f7f7f;
```

If a value can be parsed as a `Number`, instead of re-stringifying the parsed value (which will result in a decimal literal), just use the unparsed value.

### DIFF
--- a/src/io/scratch-js/toScratchJS.ts
+++ b/src/io/scratch-js/toScratchJS.ts
@@ -230,7 +230,8 @@ export default function toScratchJS(
         default:
           const asNum = Number(input.value as string);
           if (!isNaN(asNum)) {
-            return JSON.stringify(asNum);
+            // Don't re-stringify; if this can be parsed as a number, use it as-is
+            return String(input.value);
           }
           return JSON.stringify(input.value);
       }


PullJosh:
Based on the output of `npm test`, it seems like this chokes on empty strings as input:
```js
  ● sb3 -> scratch-js

    SyntaxError: Unexpected token (122:26)
      120 | yield* this.broadcastAndWait("message1");
      121 | yield* this.askAndWait(((5) + (this.random(((8) / (4)), 10))));
    > 122 | yield* this.askAndWait((() - ()));
          |                          ^
      123 | yield* this.askAndWait((() * ()));
      124 | yield* this.askAndWait((() / ()));
      125 | yield* this.askAndWait((((() < (50))) && ((() > (50)))));

      947 | 
      948 |   Object.keys(files).forEach(filepath => {
    > 949 |     files[filepath] = prettier.format(files[filepath], { ...prettierConfig, filepath });   
          |                                ^
      950 |   });
      951 | 
      952 |   return files;
```

Prettier is (correctly) upset about the empty string, which *should* be translated to a zero, instead being dropped straight into the output.

adroitwhiz:
This would probably be better + more safely done with some more advanced JS-emitting machinery that doesn't exist yet.

PullJosh:
I'm not sure I fully understand what you're saying...

adroitwhiz:
There's gotta be some library out there meant for manipulating/validating JS grammar. Right now I could do something like special-case the empty string to 0, but there could be other corner cases I'm not aware of because of the way JS gets parsed. My hope is that there's something which can deal with such scenarios.

There are [libraries](https://github.com/davidbonnet/astring) which generate a string of JS code from a syntax tree, but they assume that said syntax tree is already correct, so I don't think it'd solve this particular case.

PR #43, opened 2020-03-05
Second ESLint pass

adroitwhiz:
- Fix `npm run lint` to recurse into subdirectories
- Remove old no-longer-applicable tslint rules
- Disable problematic rules
- Make toScratchJS.ts and toScratchblocks.ts pass lint

### DIFF
--- a/src/io/scratch-js/toScratchJS.ts
+++ b/src/io/scratch-js/toScratchJS.ts
@@ -7,14 +7,14 @@ import { OpCode } from "../../OpCode";
 import * as prettier from "prettier";
 import Target from "../../Target";
 
-function triggerInitCode(script: Script) {
+function triggerInitCode(script: Script): string | null {
   const hat = script.hat;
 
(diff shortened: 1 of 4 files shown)


src/io/scratch-js/toScratchJS.ts
```diff
@@ -622,6 +624,8 @@ export default function toScratchJS(
               return `(new Date().getMinutes())`;
             case "SECOND":
               return `(new Date().getSeconds())`;
+            default:
+              return `('')`;
```
adroitwhiz: If this `default` doesn't exist, everything will fall through. An empty string matches 2.0 behavior for when this menu is set to an invalid value.

adroitwhiz:
I can do that--one sec

PullJosh:
> one sec

One... 😉

adroitwhiz:
I think any further warning fixes will require more significant changes.

PR #46, opened 2020-03-16
Add automated testing with Github Actions

### LINKED ISSUE #14: Automated testing to prevent mis-styled/etc code from being merged
It might be worth making use of something like [Travis CI](https://travis-ci.org/) (that's what LLK uses) to automatically let it lint and test code for incoming pull requests, so that we don't accidentally merge any code which could cause those to fail. Same for scratch-js!

PullJosh:
This is the [default node.js workflow](https://github.com/actions/starter-workflows/blob/master/ci/node.js.yml), unmodified. I don't think there's any way to test the workflow before merging, but I'd like to make it a PR rather than a direct commit to master so that you people can look over it first. Let me know if it seems okay!

Resolves #14


PullJosh:
We'll see how this goes!

PR #52, opened 2020-07-06
Fix handling of empty strings (fixes #51)

### LINKED ISSUE #51: Empty strings are cast to 0 in Leopard export
This means that "say ()", which is supposed to remove a speech bubble, instead causes the sprite to say 0. Currently, because of https://github.com/PullJosh/leopard/issues/66, this will remove the speech bubble, but once that is fixed, this will resurface.

PullJosh:
Should be merged in conjunction with https://github.com/PullJosh/leopard/pull/69


adroitwhiz:
Is this the intended behavior for comparison operators?

PullJosh:
Those genuinely are empty inputs in the test project, not zeroes:
image

This is more correct than the old version.

Issue #55, opened 2021-01-05
Leopard: Non-warp-mode procedures called from warp-mode procedures do not execute in warp mode


adroitwhiz:
This seems to be a limitation both in Leopard's design and in sb-edit, so I'm not sure which repo this should go in.

In Scratch, if a procedure is marked "run without screen refresh", all procedures called from within that procedure will run without refreshing the screen, even if those child procedures are not themselves marked "run without screen refresh".

In Leopard, those child procedures will always refresh the screen (yield).

A demonstration of this can be seen at https://scratch.mit.edu/projects/470160866/editor/.

PullJosh:
Aha! This is a great observation. I never fully understood what the issue was with warping in Leopard.

I don't have a super elegant solution, but one possibility is to create a `warp` function that wraps a generator and turns it into an instant script:
```js
function* mySlowProcedure(a, b) {
	for (let i = a; i <= b; i++) {
  	console.log(i);
  }
}

function warp(procedure) {
	return (...args) => {
  	const inst = procedure(...args);
    let done = false;
    while (!done) {
    	done = inst.next().done;
    }
  }
}

// Inside a "run without screen refresh" procedure, call slow procedures like this:
warp(mySlowProcedure)(1, 10);
```

Then sb-edit could add `warp` where it's needed (when calling a procedure from within a "no screen refresh" block). This doesn't feel elegant or intuitive, but it's the best idea I could come up with in 10 minutes.

adroitwhiz:
Abstracting this into a `warp` function is much cleaner than my idea. That sounds like a good solution to me!

PullJosh:
This is all fixed now and the website is updated! Thank you @adroitwhiz!

PR #56, opened 2021-01-05
Don't yield to other procedure calls inside warp-mode procedures

adroitwhiz:
This is a bit of a janky fix for #55, but I'm not sure how else to fix it. This changes the way that procedure calls are done in warp-mode procedures such that they execute the generator functions to completion, rather than yielding to them.

### DIFF
--- a/src/io/leopard/toLeopard.ts
+++ b/src/io/leopard/toLeopard.ts
@@ -766,16 +766,24 @@ export default function toLeopard(
           return `${selectedWatcherSource}.visible = true`;
         case OpCode.data_hidelist:
           return `${selectedWatcherSource}.visible = false`;
-        case OpCode.procedures_call:
-          return `yield* this.${
-            // Get name of custom block script with given PROCCODE:
-            target.scripts.find(
-              script =>
-                script.hat !== null &&
-                script.hat.opcode === OpCode.procedures_definition &&
-                script.hat.inputs.PROCCODE.value === block.inputs.PROCCODE.value
-            ).name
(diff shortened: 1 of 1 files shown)


adroitwhiz:
This will fix https://scratch.mit.edu/projects/401788041/ in https://github.com/PullJosh/leopard/issues/74

src/io/leopard/toLeopard.ts
```diff
@@ -766,16 +766,29 @@ export default function toLeopard(
...
+              script.hat.inputs.PROCCODE.value === block.inputs.PROCCODE.value
+          ).name;
+
+          const procCall = `this.${procName}(${block.inputs.INPUTS.value.map(inputToJS).join(", ")})`;
+
+          // Warp-mode procedures execute all child procedures in warp mode as well. This means running the generator
+          // functions to completion.
+          if (warp) {
+            return `{
+              const ${procName}_result = ${procCall};
```
adroitwhiz, line 784: Because generators are stateful, we execute this once to instantiate it, then call `next` on the instance. We must nest this inside a block to avoid this variable name being redeclared in case the procedure is called multiple times from within the same function.

adroitwhiz: In fact, I'll just add this in as a comment

PullJosh:
Oh! I just saw this. What do you think of https://github.com/PullJosh/sb-edit/issues/55#issuecomment-754937877? Better or worse?

adroitwhiz:
This now uses the `warp` function from https://github.com/PullJosh/leopard/pull/86.

PullJosh (approved):

PR #70, opened 2022-07-17
NPM publish process

PullJosh:
- Add Github Action for automatically publishing new releases to npm.
- Remove npm version hooks that automatically create a release on Github. (Too much automation. One fat-fingered command should not immediately create and publish an entire new version of the package.)

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,0 +1,33 @@
+# This workflow will run tests using node and then publish a package to npm when a release is created
+# For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
+
+name: Publish Node.js Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  # build:
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #     - uses: actions/checkout@v3
+  #     - uses: actions/setup-node@v3
+  #       with:
+  #         node-version: 16
+  #     - run: npm ci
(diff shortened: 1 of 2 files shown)


.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,33 @@
...
+jobs:
+  # build:
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #     - uses: actions/checkout@v3
+  #     - uses: actions/setup-node@v3
+  #       with:
+  #         node-version: 16
+  #     - run: npm ci
+  #     - run: npm test
```
adroitwhiz, line 19: Remove this commented-out code?

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,33 @@
...
+
+  publish-npm:
+    # needs: build
```
adroitwhiz, line 22: Same here

adroitwhiz (changes requested):
Looks good as far as I can tell (I usually need to try a few times before GitHub runs my workflows properly). It might be a good idea to remove the commented-out parts, though.

.github/workflows/npm-publish.yml
PullJosh: Actually... I removed this because that's what I did for `leopard`, but actually we have test cases for `sb-edit`, so we might as well keep this part of the action.

adroitwhiz (approved):

PR #74, opened 2022-07-19
Add support for "when greater than" hat blocks

adroitwhiz:
I had to make a few changes to get this to work:

- `triggerInitCode` is now inside `toLeopard`.
- The `blockToJS` and `scriptToJS` code has been moved inside a new function, `blockToJSWithContext`, which both `scriptToJS` and `triggerInitCode` call. It wraps `blockToJS` and `scriptToJS` with the requisite `script` (now optional) and `target`.
- `triggerInitStr` now expects callers to stringify their own option values, since they can now be functions as well, and those can't be `JSON.stringify`'d.


adroitwhiz:
Also, it looks like I didn't have to pass `target` into `Trigger` methods at all--the `this` value in the arrow function is automatically bound to the `Sprite` in which the trigger is defined!

src/io/leopard/toLeopard.ts
```diff
@@ -198,8 +171,66 @@ export default function toLeopard(
     }
   }
 
+  function staticBlockInputToLiteral(value: string | number | boolean | object): string {
+    const asNum = Number(value as string);
+    console.log(value, asNum);
```
PullJosh, line 176: Remove `console.log`?

PullJosh (approved):
Looks good! 😄

Thank you so much for making all these improvements.

PR #89, opened 2023-02-17
Reference/resolve variables by ID instead of name

### LINKED ISSUE #85: Cannot read properties of undefined (Reading "0")
Title, this error is causing me hell when trying to convert a friend's project using Leopard.
I tracked the error down to an IO library in the converter. The exact error spot is sb-edit/src/io/leopard/toLeopard.js:927.
Full error (Glitch.me, running project Colorlens by Yuski347):

/rbd/pnpm-volume/c3f6456c-ea9a-4017-ae1c-18ff3b9a1c5b/node_modules/sb-edit/lib/io/leopard/toLeopard.js:927
                })[0]
                  ^
TypeError: Cannot read properties of undefined (reading '0')
    at /rbd/pnpm-volume/c3f6456c-ea9a-4017-ae1c-18ff3b9a1c5b/node_modules/sb-edit/lib/io/leopard/toLeopard.js:927:19
    at Array.map (<anonymous>)
    at _loop_1 (/rbd/pnpm-volume/c3f6456c-ea9a-4017-ae1c-18ff3b9a1c5b/node_modules/sb-edit/lib/io/leopard/toLeopard.js:921:93)
    at Project.toLeopard (/rbd/pnpm-volume/c3f6456c-ea9a-4017-ae1c-18ff3b9a1c5b/node_modules/sb-edit/lib/io/leopard/toLeopard.js:949:9)

### LINKED ISSUE #86: toLeopard: Watchers which are only referred to in a "hide variable/list" block don't get generated
Sure, these are technically no-ops, but in the interests of having Leopard/sb-edit not "eat" blocks, it'd be better to define these watchers than just delete associated "hide variable/list" blocks.

As-is we get such errors as:

```js
*whenIReceiveUnDev() {
  this.watchers.infiniteHealth.visible = false;
  this.watchers.absoluteChaos.visible = false;
  this.watchers.infiniteEx.visible = false; // Uncaught TypeError: Cannot set properties of undefined
  this.watchers.infiniStomp.visible = false;
}
```

towerofnix:
**This is a major change to the way sb-edit reads and serializes variables.**

- Fixes #85 
- Fixes #86 

This PR changes `BlockInput.Variable` and `BlockInput.List` so their value is a pair of values (in an object), `{ id: string, name: string }`, instead of one value (just the name). The `fromSb3` deserializing code pulls from the IDs in the actual sb3 file, and all serializing code is updated to respect the new format — most notably, this resolves issues in `toLeopard` surrounding multiple variables existing of different IDs but the same name.

The last commit, af5280ba271a184aa0a218ab7b306a7f2f31377c, adds a final pass to `fromSb3` deserialization so that variables which are never referenced in blocks nor shown as monitors are totally wiped from the code — this is a bit of a presumptuous decision and means sb3 -> sb3 conversion is explicitly *not* a one-to-one conversion, but I think that's an okay precedent to set, and it seemed to make more sense to run this pass on input data in `fromSb3` rather than preprocessing in `toLeopard`. (In the future, I'm considering separating this chunk of code out so it's one of a modular set of "processing" prefabs including other automatic code optimizations, but that would be for a separate PR.)

@adroitwhiz I added you as a PR reviewer, but feel free to pass if you aren't interested or don't currently have the time!

### DIFF
--- a/src/io/leopard/toLeopard.ts
+++ b/src/io/leopard/toLeopard.ts
@@ -99,7 +99,7 @@ export default function toLeopard(
 
   let targetNameMap = {};
   let customBlockArgNameMap: Map<Script, { [key: string]: string }> = new Map();
-  let variableNameMap: Map<Target, { [key: string]: string }> = new Map();
+  let variableNameMap: { [id: string]: string } = {}; // ID to unique (Leopard) name
 
   for (const target of [project.stage, ...project.sprites]) {
     const newTargetName = uniqueName(camelCase(target.name, true));
@@ -108,19 +108,9 @@ export default function toLeopard(
 
     let uniqueVariableName = uniqueName.branch();
 
-    const varNameMap = {};
-    variableNameMap.set(target, varNameMap);
-
-    for (const list of target.lists) {
-      const newName = uniqueVariableName(camelCase(list.name));
-      varNameMap[list.name] = newName;
-      list.setName(newName);
-    }
-
-    for (const variable of target.variables) {
-      const newName = uniqueVariableName(camelCase(variable.name));
-      varNameMap[variable.name] = newName;
-      variable.setName(newName);
+    for (const { id, name } of [...target.lists, ...target.variables]) {
+      const newName = uniqueVariableName(camelCase(name));
+      variableNameMap[id] = newName;
     }
 
     const uniqueScriptName = uniqueNameGenerator([
@@ -276,24 +266,25 @@ export default function toLeopard(
 
       // If the block contains a variable or list dropdown,
       // get the code to grab that variable now for convenience
-      let varName: string = null;
       let selectedVarSource: string = null;
       let selectedWatcherSource: string = null;
+      let varInputId: string = null;
+      let isSpriteVar: boolean = null;
       if ("VARIABLE" in block.inputs) {
-        varName = block.inputs.VARIABLE.value.toString();
+        varInputId = (block.inputs.VARIABLE.value as { id: string }).id;
+        isSpriteVar = target.variables.some(({ id }) => id === varInputId);
+      } else if ("LIST" in block.inputs) {
+        varInputId = (block.inputs.LIST.value as { id: string }).id;
+        isSpriteVar = target.lists.some(({ id }) => id === varInputId);
       }
-      if ("LIST" in block.inputs) {
-        varName = block.inputs.LIST.value.toString();
-      }
-      if (varName !== null) {
-        const spriteVars = variableNameMap.get(target);
-        if (varName in spriteVars) {
-          selectedVarSource = `this.vars.${spriteVars[varName]}`;
-          selectedWatcherSource = `this.watchers.${spriteVars[varName]}`;
+      if (varInputId) {
+        const newName = variableNameMap[varInputId];
+        if (isSpriteVar) {
+          selectedVarSource = `this.vars.${newName}`;
+          selectedWatcherSource = `this.watchers.${newName}`;
         } else {
-          const stageVars = variableNameMap.get(project.stage);
-          selectedVarSource = `this.stage.vars.${stageVars[varName]}`;
-          selectedWatcherSource = `this.stage.watchers.${stageVars[varName]}`;
+          selectedVarSource = `this.stage.vars.${newName}`;
+          selectedWatcherSource = `this.stage.watchers.${newName}`;
         }
       }
 
@@ -643,7 +634,10 @@ export default function toLeopard(
               if (block.inputs.OBJECT.value !== "_stage_") {
                 varOwner = project.sprites.find(sprite => sprite.name === targetNameMap[block.inputs.OBJECT.value]);
               }
-              propName = `vars[${JSON.stringify(variableNameMap.get(varOwner)[block.inputs.PROPERTY.value])}]`;
+              // "of" block gets variables by name, not ID, using lookupVariableByNameAndType in scratch-vm.
+              const variable = varOwner.variables.find(variable => variable.name === block.inputs.PROPERTY.value);
+              const newName = variableNameMap[variable.id];
+              propName = `vars.${newName}`;
               break;
             }
           }
@@ -976,7 +970,7 @@ export default function toLeopard(
     // Some watchers start invisible but appear later, so this code builds a list of
     // watchers that appear in "show variable" and "show list" blocks. The list is
     // actually *used* later, by some other code.
-    let shownWatchers = new Set();
+    let shownWatchers: Set<string> = new Set();
     let targetsToCheckForShowBlocks: Target[];
     if (target.isStage) {
       targetsToCheckForShowBlocks = [project.stage, ...project.sprites];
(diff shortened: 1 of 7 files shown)


towerofnix:
Will need to regen snapshots for this PR and the follow-up one, rn I am crashing for the night! 🐌

PullJosh:
Hey! I have been looking at this code in my moments of free time... It will take me a bit longer to review everything, but I just wanted to send you a message to say a huge thanks for the contribution. 😄

I really appreciate you continuing to contribute fixes! ❤️

towerofnix:
No problem at all! I've been enjoying popping back into this codebase. Leopard is really one of my favorite projects surrounding Scratch, so I'm glad to have the opportunity to poke around the code and improve it! ✨

PullJosh (approved):
Looks good to me! :) I left two comments that are worth reading, but nothing that blocks a merge in my opinion.

@towerofnix I'll leave it up to you to actually merge the PR because you have multiple PRs open and I don't know which order will be easiest. (Once all the changes have been merged, I am happy to handle publishing the updates to npm and redeploying the Leopard website.)

src/io/leopard/toLeopard.ts
```diff
@@ -276,24 +266,25 @@ export default function toLeopard(
...
       // If the block contains a variable or list dropdown,
       // get the code to grab that variable now for convenience
-      let varName: string = null;
       let selectedVarSource: string = null;
       let selectedWatcherSource: string = null;
+      let varInputId: string = null;
+      let isSpriteVar: boolean = null;
       if ("VARIABLE" in block.inputs) {
-        varName = block.inputs.VARIABLE.value.toString();
+        varInputId = (block.inputs.VARIABLE.value as { id: string }).id;
```
adroitwhiz, line 274: The type assertion here seems to be because the block's `VARIABLE` input can only be narrowed to `BlockInput.Any`. Something to look into in a future PR?

src/io/sb3/fromSb3.ts
```diff
@@ -423,6 +427,70 @@ export async function fromSb3JSON(json: sb3.ProjectJSON, options: { getAsset: Ge
...
+        if (variable.visible) {
+          continue;
+        }
+        if (usedVariableIds.has(variable.id)) {
+          continue;
+        }
+
+        varList.splice(i, 1);
+        i--;
+        continue;
```
adroitwhiz, line 471: This `continue` appears at the end of the loop, so it can be removed, right?

src/io/sb3/fromSb3.ts
```diff
@@ -423,6 +427,70 @@ export async function fromSb3JSON(json: sb3.ProjectJSON, options: { getAsset: Ge
     videoOn: stage.videoState === "on",
     videoAlpha: stage.videoTransparency
   });
+
+  const targetToBlocks: Map<Target, Array<Block>> = new Map();
+  for (const target of [project.stage, ...project.sprites]) {
+    targetToBlocks.set(target, flattenBlocks(target.scripts.flatMap(script => script.blocks)));
```
adroitwhiz, line 433: What does `flatMap` do here? Does it concatenate every block in every script into one array?

src/io/sb3/fromSb3.ts
```diff
@@ -423,6 +427,70 @@ export async function fromSb3JSON(json: sb3.ProjectJSON, options: { getAsset: Ge
...
+    }
+  }
+
+  return project;
+}
+
+function flattenBlocks(blocks: Array<Block>) {
+  return blocks.flatMap(block => [
+    block,
+    ...flattenBlocks(
```
adroitwhiz, line 482: How many intermediate arrays are we creating here? I'm worried this could be a performance issue for nested blocks, but on the other hand I don't expect the average Scratch project to next blocks that deeply.

src/io/sb3/fromSb3.ts
adroitwhiz: I'm wondering if this could be folded into `flattenBlocks`-- a `Script`'s `blocks` is a `Block[]`, and so is a C-block's substack(s).

src/io/sb3/fromSb3.ts
adroitwhiz: For instance, could we pass the final block array into `flattenBlocks` and then append to it within each recursive call?

src/io/leopard/toLeopard.ts
```diff
@@ -1051,32 +1049,29 @@ export default function toLeopard(
...
+                `this.vars.${variableNameMap[variable.id]} = ${toOptimalJavascriptRepresentation(variable.value)};`
             )
-            .filter(([variable, oldName]) => variable.visible || shownWatchers.has(oldName))
-            .map(([variable, oldName]) => {
-              return `this.watchers.${variable.name} = new Watcher({
-              label: ${JSON.stringify((target.isStage ? "" : `${target.name}: `) + oldName)},
+            .join("\n")}
+
+          ${[...target.variables, ...target.lists]
+            .map(variable => [variable, variableNameMap[variable.id]] as [Variable | List, string])
```
adroitwhiz, line 1060: Why this `map` here? The variable's name isn't used in the subsequent `filter` operation, so it looks like fetching the variable's name from `variableNameMap` could be moved into the function after the `filter`.

adroitwhiz (changes requested):
The logic looks good overall. My main concern is that the code is written in a style that's a bit *too* functional for JS in my opinion. While functional programming languages like Haskell [optimize chained map/filter/etc. operations on lists](https://wiki.haskell.org/GHC_optimisations#Fusion), JS engines will happily create every intermediate array eagerly.

My other nitpick (not addressed in the individual review comments) is that the existing codebase uses `T[]` to represent arrays, while this PR uses `Array<T>`; it's probably better to match the existing coding style.

Also, it looks like Prettier hasn't been run over the code yet (`npm run format`).

src/io/leopard/toLeopard.ts
```diff
@@ -986,11 +980,15 @@ export default function toLeopard(
...
+            shownWatchers.add(
+              (block as BlockBase<OpCode.data_showvariable, { VARIABLE: BlockInput.Variable }>).inputs.VARIABLE.value.id
+            );
           }
-          if (block.opcode === OpCode.data_showlist) {
-            shownWatchers.add(block.inputs.LIST.value);
+          if (block.opcode === OpCode.data_showlist || block.opcode === OpCode.data_hidelist) {
+            shownWatchers.add(
+              (block as BlockBase<OpCode.data_showlist, { LIST: BlockInput.List }>).inputs.LIST.value.id
+            );
```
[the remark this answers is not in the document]
towerofnix: Thanks for the catch! There were errors here earlier during testing that I couldn't get TypeScript to avoid without this. I'm not sure what was causing it since this code does work without the assertions, as you said.

src/io/sb3/fromSb3.ts
towerofnix: Oops, this is an issue I addressed in 560096ee42fec48fdea4f2ac1a2ecd9e0e55ba2b. I meant to include that on this pull request but forgot.

src/io/leopard/toLeopard.ts
towerofnix: Thanks, good catch. Previous logic had the following `filter` make use of the name (previously the *old* i.e. Scratch name), for checking in the "shown watchers" list. Your suggestion makes the intermediate array unneeded, which is a nice optimization (both in performance and conceptually)!

src/io/leopard/toLeopard.ts
towerofnix: Yep, marked this on your issue #92. Thanks!

towerofnix:
Thanks, both! I've gone over your code reviews. The most notable change is in the implementation for `Target.blocks` — a property which I updated for this purpose but forgot to push to this branch! — matching your suggestion of recursively pushing to an "all blocks" collector array instead of spreading/flat-mapping intermediates, @adroitwhiz.

I've also made similar changes and optimizations throughout, see my recent commits. Most notable: [caching a set of stage variables](https://github.com/leopard-js/sb-edit/pull/89/commits/36f093413e18a099f8d186e74df377a682cdfe0d) instead of iterating over `target.variables` every time, as well as entirely short-circuiting the set check if the target is the stage.

adroitwhiz (approved):
Looks good now! I didn't even realize `Target.blocks` existed.

towerofnix:
@adroitwhiz Yep, it wasn't visible when you reviewed because I pushed it to `compat` (which is based off this branch) and forgot to push it here too. 📦 

Thanks both for the reviews!

@PullJosh I've merged this pull request. It's not dependent on any other PRs, as my other pull requests are either independent or support for #90. They should all be reviewed, but here's a summary of their relationships to help guide you:

* Major pull request: #90
  * Supported by leopard-js/leopard#159
  * Supported by leopard-js/leopard#161
  * Also supported by this pull request. Now that it's been merged, it will be easier to review the changes which were introduced in sb-edit#90.
* Independent pull request: #91
  * No user-visible changes, only for alternate consumers of sb-edit e.g. [sb-edit-playground](https://github.com/towerofnix/sb-edit-playground).
* Independent pull request: leopard-js/leopard#160
  * Fixes user-visible bugs, i.e. when setting costume to 0 or below (which wraps back from the last costume in Scratch). This fix doesn't require any changes in sb-edit.

Since this pull request fixes user-visible bugs (failing to import some projects such as seen in #85 and leopard-js/leopard#162) and does not depend on any other pull requests, it's ready and useful to end users to release a version of sb-edit before other pull requests are reviewed.

PR #90, opened 2023-02-17
Compatibility changes & Leopard serialization polish

### LINKED ISSUE #60: Does not rename objects with the same name as an import
I have a Scratch project with object "Sound" which interferes with the import "Sound".

```js
import {
  Sprite,
  Trigger,
  Watcher,
  Costume,
  Color,
  Sound
} from "https://unpkg.com/leopard@^1/dist/index.esm.js";

export default class Sound extends Sprite {
//                   ^^^^^ Uncaught SyntaxError: Identifier 'Sound' has already been declared
```

### LINKED ISSUE #87: toLeopard: "false" != false, "true" != true
`"false"` is a truthy value, and the non-strict equality comparison says it isn't equal to `false` (the boolean value). This makes runtime comparisons between false and "false" fail.

This is apparently also the case for `"true"`/`true` (`"true" != true). Heavens know why, I don't!

### LINKED ISSUE #88: toLeopard: Non-boolean blocks should be cast to booleans in boolean inputs
For whatever reason, Scratch lets you put the "item (index) of (list)" block into boolean inputs like "if":

set i to 1; repeat length of inputs; if item i of inputs then...

These need to be cast to boolean.

towerofnix:
This pull request does... a whole lot!!! Sorry, not sorry, I usually don't do monolithic PRs but sometimes you gotta take what you can get. 👻 

**This PR builds off of #89.** To isolate this PR's changes before that PR is merged: https://github.com/towerofnix/sb-edit/compare/var-ids...towerofnix:sb-edit:compat

* Continues from #89
* Depends on leopard-js/leopard#157
* Depends on leopard-js/leopard#159
* Depends on leopard-js/leopard#161
* Fixes #60
* Fixes #87
* Fixes #88
* Fixes leopard-js/leopard#156
* Resolves leopard-js/leopard#58
* A whole host of other changes and legibility improvements, see below!

Major changes summarized:

* New `desiredInputShape` feature exists so that inputs are casted as needed (improves Scratch compatibility) and can be mutated from within `blockToJs`, providing custom equivalent-behavior forms to improve code legibility.
  * Possible values: `any`, `index`, `number`, `string`, `boolean`, `stack` (all strings)
  * The block-to-source map is refactored to set a new `blockSource` variable instead of returning right away.
  * Each entry in block-to-source map also sets the kind of input shape it naturally satisfies, e.g. math blocks always satisfy `"number"`, conditionals always satisfy `"boolean"`.
  * If a block satisfies the desired input type, its source text is returned as-is. Otherwise, it's wrapped with a casting function.
  * Special handling for `index` desired input shape (which is for use in list and string indexing):
    * Works by changing `satisfiesInputShape` to `index` (instead of `number`) if it was specially handled
    * **`operator_add`:** If one of the values is a number and not a block (right-hand preferred), it is decremented by one, e.g. `this.vars.list[this.vars.i + 1]` instead of `this.vars.list[this.vars.i + 2 - 1]`; if the decremented value is just zero, returns the other side as-is, e.g. `this.vars.list[this.vars.i]` instead of `this.vars.i + 1 - 1]`
    * **`data_itemnumoflist`:** Does not do `+ 1` if the desired input shape is index, e.g. `this.vars.list[this.vars.indexOf(this.vars.foo)] = 42` instead of `this.vars.list[this.vars.indexOf(this.vars.foo) + 1 - 1] = 42`
    * **Generic numbers:** Decremented by 1 if the desired input shape is index, e.g. `this.vars.list[0]` instead of `this.vars.list[1 - 1]`
  * Some inputs may be more stringent than necessary for Leopard's purposes, e.g. it's possible `this.goto` and `this.glide` will work even if the input hasn't been casted to number, and so on. This could use some more investigation.
* Use new `this.compare()` Leopard-provided function for comparisons when needed
  * Analogous to scratch-vm's `Cast.compare` function, implementation based on @apple502j's optimized code: https://github.com/LLK/scratch-vm/pull/2580
  * Aiming for compatibility, this is always used for greater-than and less-than. This can probably be avoided when comparing against strings, but Scratch will treat numbers *as strings* (for the sake of comparison) if either side can't be parsed as a number, which makes using `>` or `<` for numbers awkward.
  * Easier to get away for equality comparisons: if either side is a static string or number, we cast the other side to the appropriate type and perform a strict equality check (`===`). I'm not sure this is exactly correct 100.00% of the time (particularly if Infinity is involved), but it should cover the vast majority of use cases.
  * Future work: `x === y` could be converted to `x !== y` instead of `!(x === y)`
* Use new `increase()` and `decrease()` "macros" for common operations, so that `++`, `--`, `+=`, and `-=` are all used as appropriate, instead of always using `+=`
  * `++` and `--` are disabled in most instances (only used for `data_setvariableto` and next costume/backdrop blocks), since they don't really fit the "feel" of changing tangible values — this matches existing Leopard behavior while allowing for `this.vars.i++`, which should feel more appropriate when treating variables as abstract data
* Remove "branch" behavior from `uniqueNameGenerator` and just spawn totally new generators when it makes sense
  * This fixes a serious legibility issue where sprite-local variables (and custom block arguments!) were being generated *with unique names across the entire project*, e.g. `this.vars.i`, `this.vars.i2` ... `this.vars.i12`, `this.vars.i13` if you had 13 sprites with variables named `i`!
  * There didn't seem to be any issues resulting from this change — we store generated unique names on a project-wide map, so accessing a variable from another sprite (with `sensing_of`) still works fine. There's no possibility for names to be duplicated within the same namespace!
  * I think this is what the "branch" behavior was probably supposed to do from the start, but it didn't actually accomplish that. It's simpler to just create fresh name generators when needed.
* Miscellaneous compatibility changes
  * `data_itemoflist` and `operator_letter_of` use the new helper functions to return an empty string when the index is out of range
  * `operator_contains`, `data_itemnumoflist`, and `data_listcontainsitem` use new helper functions for compatibility (parity with scratch-vm `Cast.compare`, most notably these are all case-insensitive)
  * `operator_mathop` uses new helper function for "tan" operation, Scratch has a custom implementation
  * Reserved name lists have been greatly expanded, newly including reserved names at the module/global-level too (e.g. `Costume`, `Sprite`) — sprite names can't conflict with these!
    * When identifiers aren't on a namespace (like `this` or `this.vars`), we also include a list of reserved JavaScript words — this avoids issues with custom block arguments with JS-reserved names such as `function` or `class`.
  * `Target.blocks` was dubiously "flat" before, but only one level deep (e.g. "repeat 10" under a "forever" loop, but not "turn 15 degrees" under that "repeat 10"). This has been fixed, which is a sb-edit format change, but since this property seemed to be intended to flatten scripts in the first place, it seems more like a bug fix than a breaking interface change.
    * This fixes nested "show/hide variable/list" blocks being missed when determining which variables have watchers that get dynamically shown/hidden at runtime.
  * For variable watchers, slider min/max and `step` are now passed through the watcher constructor.
    * `step` is 1 if "discrete", 0.01 otherwise. Scratch's default is discrete, only non-discrete if at least one of the bounds is written as a decimal e.g. `1.00` — this is conveniently serialized info scratch-vm makes directly available instead of having to parse sliderMin/sliderMax ourselves (which is a number, not a string, anyway).

This is a pretty big PR, but by covering a lot of ground, it should go a long way to make the output of Leopard more legible and usable as an actual basis for starting new projects from. This is what you get when I'm recovering from a cold and don't have anything better to do ^_^ (Except... sleep? It's almost 1AM? Oh, I'll hear none of that! 💤)

Same as in the other PR: @adroitwhiz I've added you as a reviewer *if you'd like to!* If you're not interested or don't currently have the time, feel free to pass. 👾✨


adroitwhiz:
I'll try to take a look at this later today. Would you mind rebasing it on top of the main branch since #89 is merged?

towerofnix:
Rebase done!

src/io/leopard/toLeopard.ts
```diff
@@ -1,14 +1,75 @@
...
+];
+
+type InputShape = "any" | "index" | "number" | "string" | "boolean" | "stack";
+
 function uniqueNameGenerator(usedNames: string[] = []) {
+  usedNames = usedNames.slice();
+  return uniqueName;
+
   function uniqueName(name): string {
     if (!usedNames.includes(name)) {
```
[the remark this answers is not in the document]
towerofnix: Good catch!

src/io/leopard/toLeopard.ts
```diff
@@ -8,10 +8,71 @@ import * as prettier from "prettier";
 import Target from "../../Target";
 import { List, Variable } from "../../Data";
 
-function uniqueNameGenerator(usedNames: string[] = []) {
+// Words which are invalid for any JavaScript identifier to be, when it isn't
+// on a namespace (like `this` or `this.vars`).
+//
+// This list may be more comprehensive than it needs to be in every case,
+// erring to avoid potential issues.
```
adroitwhiz, line 15: It might be worth adding a comment detailing where this list of reserved words comes from (MDN? the spec?)

src/io/leopard/toLeopard.ts
```diff
@@ -8,10 +8,71 @@ import * as prettier from "prettier";
...
+  "try",
+  "typeof",
+  "yield",
+  "var",
+  "void",
+  "while",
+  "with"
+];
+
+type InputShape = "any" | "index" | "number" | "string" | "boolean" | "stack";
```
adroitwhiz, line 67: It might be worth using a TypeScript `enum` here

src/io/leopard/toLeopard.ts
```diff
@@ -253,7 +422,41 @@ export default function toLeopard(
...
+      } else if (allowIncrementDecrement && n === -1) {
+        return `${leftSide}--;`;
+      } else if (n >= 0) {
+        return `${leftSide} += ${JSON.stringify(n)};`;
+      } else if (n < 0) {
+        return `${leftSide} -= ${JSON.stringify(-n)};`;
+      }
+    }
+
+    function decrease(leftSide: string, input: BlockInput.Any, allowIncrementDecrement: Boolean = true) {
```
adroitwhiz, line 442: A few concerns here:
- How come `decrease` defaults `allowIncrementDecrement` to `true` but `increase` doesn't? When would we want to disallow increment/decrement operators?
- There's a lot of shared logic between `increase` and `decrease`; is it possible to combine them?
- In TypeScript, the capital-B `Boolean` [refers to a boxed `Boolean` object, which is probably not what we want for the function signature](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#number-string-boolean-symbol-and-object).

src/io/leopard/toLeopard.ts
```diff
@@ -253,7 +422,41 @@ export default function toLeopard(
...
+      let n;
+      if (input.type === "block" || ((n = Number(input.value)), isNaN(n))) {
```
adroitwhiz, line 444: How many places are we doing an "input is a primitive, and its value is X type" check? It might be worth abstracting it into a function.

adroitwhiz (changes requested):
Did a quick review; didn't have time to go over all the individual block code generation changes yet. My main question would be: is it possible to abstract "check whether this block input is a literal with X type" into a function? It looks like we're relying on that operation to elide casts in a *lot* of places, and there's a lot of duplicate code.

src/io/leopard/toLeopard.ts
towerofnix: > How come decrease defaults allowIncrementDecrement to true but increase doesn't?

That's an error. Thanks for the catch.

> There's a lot of shared logic between increase and decrease; is it possible to combine them?

There is, but these aren't conceptually heavy functions. This is a code style opinion: if two functions have little to gain from being merged into one — and would necessitate a comparatively significant amount of boilerplate — then it is probably better to leave them as separate. 

<details><summary>Long commentary on this subject</summary>
The overlap between the two functions is:

* The actual syntax tokens, i.e. `x++`, `x--`, `x += stringify(n)`, `x -= stringify(n)`
* The overall logical structure, i.e. fallback for blocks & non-number primitives then four math checks to determine syntax
* The specifics of the first step, i.e. how it detects if the input is a block or non-number primitive

The difference between the two functions is:

* When those syntax tokens are used
* The details of the third and fourth conditions (`n >= 0`, `n < 0` vs `n > 0`, `n <= 0`)
* The syntax token returned on block or non-number primitive fallback

If we tried to merge the two, we would have to face these challenges:

* `increase` and `decrease` are fundamentally distinct modes. We would have to represent which mode is desired, which requires passing an arbitrary string or enum value, adding boilerplate within the function and when calling it.
* Within the function, we would have to either find a clean way to "interweave" the associations between conditions and syntax tokens for increase/decrease, which would likely increase logical complexity, or have entirely separate branches for increase/decrease, which would add boilerplate. Either way leaves you with a function whose two second halves are functionally independent of each other and an arbitrary, non-idiomatic "switch" to decide which mode is preferred.

It is possible to move the first half of the function (fallback detection) into its own function, which would be helpful in theory since it's shared between the two. However, we actually care about two results from this first half of the code:

1. Whether or not the input was primitive *and* could be successfully parsed as a number
2. If it could be parsed as a number, what the number was

This is *possible* to represent with a single function, but requires boilerplate (and allocating more memory for multiple return values from a single function, as a short-lived array or object, if V8 doesn't optimize that out <sup>[* I touch on this again later]</sup>).

This isn't something you brought up, but the syntax I ended up using was a little awkward but was the simplest representation I could figure out, and compact. An alternative would be using an extra variable.

```ts
let n, returnCast;
if (input.type === "block") {
  returnCast = true;
} else {
  n = Number(input.value);
  if (isNaN(n)) {
    returnCast = true;
  }
}

if (returnCast) {
  return `${leftSide} ...`;
}

if (allowIncrementDecrement && n === 1) ...
```

In a separate function (meaning it's only written once), this isn't so bad, albeit a bit clunky. However, it still runs into the issue of returning both `returnCast` and `n`. Moreover, the function would only ever be used in `increase` or `decrease` and so make you wonder if it ought to never have been given a reusable form.

An alternative is to use a helper function which covers the entire structure of the function, using an anonymous callback to handle the case where the input was parsed as a number. This solves most of the problems brought up because it allows `increase` and `decrease` to still exist while not covering more logic than they strictly need to. A helper function is explicitly intended for its limited use cases, so is easier to justify than a limited-use-case "check if X can be parsed *and* see what it was parsed as" type function.(*)

```ts
function increase(leftSide, input, allowIncrementDecrement) {
  return increaseDecreaseHelper(input,
    () => `${leftSide} += (${inputToJS(input, "number")});`,
    n => {
      if (allowIncrementDecrement && n === 1) ...
    });
  // same for decrease
}

function increaseDecreaseHelper(input, handleFallback, handleNumberPrimitive) {
  let n;
  if (/* logic */) {
    return handleFallback();
  } else {
    return handleNumberPrimitive(n);
  }
}
```

(*) The actual footprint of that function could be reduced by making it return `null` if it fails to parse. This would be a "better looking" way of having two different modes of returning ("yes, it parsed, here is the value" and "no, it didn't parse, enter a different branch of code"). However, it still necessitates boilerplate for the caller (check if parsed === null, fallback if so) and doesn't change the real conceptual complexity of the function.

But the helper function still introduces boilerplate (remember the exact form of `increaseDecreaseHelper`... sure, it's just used twice, but STILL...) and just kind of looks bad.

---

The whole theme of this comment is considering the complexity *introduced by* abstracting, and whether or not hiding some other complexity is worth the added overall complexity.

One way to justify a general-purpose function with a specific arbitrary form is by having two similar functions, `parseAsString` and `parseAsNumber`, which justify the conceptual complexity by sharing the same "language" (form).

Another is just by reusing it enough. If the form is repeated enough, it gains legitimacy because the abstract form is easier to memorize than "syncing" identical changes across each place the identical logic wasn't abstracted.

Another is by the logic it replaces being sufficiently burdensome that it's worth memorizing a form to abstract it. It doesn't have to be repeated as many times to justify its existence because it's also doing the work of cutting out something that is more awkward to repeat than the form replacing it, even though it is less abstract.

I don't think we have any instances of a `parseAsString` function being useful, but there are several where the logic that would be abstracted in `parseAsNumber` is repeated. Also, it's a little awkward to repeat `(block.inputs.NUM2 as BlockInput.Any).type !== "block" && !isNaN(Number(block.inputs.NUM2.value))` every time - moreover in that case we don't even reuse `Number(block.inputs.NUM2.value)`, so that's *less* optimized code that would freely be optimized by being changed to fit the abstract form.
</details>

**TL;DR:** It's probably not worth it to combine `increase` and `decrease` or try to fully consolidate the shared logic into a helper function, but a `parseAsNumber` function which returns null for unparseables would be useful in other instances as well, and would do most of the work a more specialized helper function would anyway.

src/io/leopard/toLeopard.ts
towerofnix: Good point.

src/io/leopard/toLeopard.ts
```diff
@@ -663,149 +1155,375 @@ export default function toLeopard(
...
         case OpCode.operator_multiply:
-          return `((${inputToJS(block.inputs.NUM1)}) * (${inputToJS(block.inputs.NUM2)}))`;
+          satisfiesInputShape = "number";
+          blockSource = `((${inputToJS(block.inputs.NUM1, "any")}) * (${inputToJS(block.inputs.NUM2, "any")}))`;
+          break;
+
         case OpCode.operator_divide:
-          return `((${inputToJS(block.inputs.NUM1)}) / (${inputToJS(block.inputs.NUM2)}))`;
+          satisfiesInputShape = "number";
+          blockSource = `((${inputToJS(block.inputs.NUM1, "any")}) / (${inputToJS(block.inputs.NUM2, "any")}))`;
```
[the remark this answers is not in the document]
towerofnix: JS automatically numberalizes values for these operations, but to be fair it also stringifies the second input in `"string" + any` so I'll change these to `"number"` for consistency.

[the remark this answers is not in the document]
towerofnix: Thanks, good catch. Didn't think about what would happen when you gave it something that couldn't be a number in the first place 😅

src/io/leopard/toLeopard.ts
```diff
@@ -663,149 +1155,375 @@ export default function toLeopard(
...
+          blockSource = `this.random(${inputToJS(block.inputs.FROM, "any")}, ${inputToJS(block.inputs.TO, "any")})`;
+          break;
+
         case OpCode.operator_gt:
-          return `((${inputToJS(block.inputs.OPERAND1)}) > (${inputToJS(block.inputs.OPERAND2)}))`;
+          satisfiesInputShape = "boolean";
+          blockSource = `(this.compare((${inputToJS(block.inputs.OPERAND1, "any")}), (${inputToJS(
+            block.inputs.OPERAND2,
+            "any"
+          )})) > 0)`;
```
towerofnix, line 1263: Worth investigating if it's possible to skip `this.compare` in some cases here, particularly if one of the inputs is a number.

src/io/leopard/toLeopard.ts
```diff
@@ -663,149 +1155,377 @@ export default function toLeopard(
...
+            }
+          }
+          satisfiesInputShape = "number";
+          blockSource = `((${inputToJS(block.inputs.NUM1, "number")}) + (${inputToJS(block.inputs.NUM2, "number")}))`;
+          break;
+
         case OpCode.operator_subtract:
-          return `((${inputToJS(block.inputs.NUM1)}) - (${inputToJS(block.inputs.NUM2)}))`;
+          satisfiesInputShape = "number";
+          blockSource = `((${inputToJS(block.inputs.NUM1, "number")}) - (${inputToJS(block.inputs.NUM2, "number")}))`;
```
towerofnix, line 1240: Should be updated to "absorb" following decrement, like `operator_plus` above.

adroitwhiz (approved):
Looks good! In the future, it may be worth looking into whether `inputToJS` could return the type of the JS expression it returns, but that's definitely outside the scope of this PR.

PullJosh (approved):

towerofnix:
All PR's approved so this is good to go! 🚀

PR #127, opened 2024-03-03
toLeopard: Add support for "if on edge, bounce"

towerofnix:
Self-explanatory. This is the function defined in leopard-js/leopard#188 / leopard-js/leopard#114 and should be merged/released after that pull request.


adroitwhiz (approved):
Good to go!

PR #129, opened 2024-04-23
Create CLI that can convert sb3 to leopard or leopard-zip

### LINKED ISSUE #128: Command line tool
The Leopard website is a nice way to convert projects, but it's clearly valuable to give people the power to use sb-edit locally. And they shouldn't have to write JavaScript code in order to do it. ([Here's an example](https://scratch.mit.edu/discuss/topic/420162/?page=50#post-7866901) of someone trying to convert from sb3 to Leopard locally who is struggling.)

It feels sort of obvious to have a command-line utility that allows people to run `npx sb-edit [command]` and get all kinds of useful outputs.

I'm not a big CLI user so I don't have a good intuition for what the commands/options should be. But the obvious main feature is to **let users choose an input project and an output format/destination and then convert the input to output automatically**.

Input options:
* Scratch project URL
* .sb3 file

Output options:
* Leopard project (.zip or just a plain directory)
* .sb3 file
* Scratchblocks output

**If anybody is interested in taking this project on, that would be fantastic!**

PullJosh:
Closes #128. **Allows importing a .sb3 file and generating a leopard folder or zip directly from the CLI.** I haven't implemented .sb3 output yet for the sake of time/effort. (And because .sb3 is currently the only allowed input method so it would be mostly useless.)

Example output:
Screenshot of sb-edit CLI output

I also did my best to make the errors look nice:
Screenshot of sb-edit CLI error

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -1,0 +1,248 @@
+#! /usr/bin/env node
+
+import { Command, Option } from "commander";
+import Project from "../Project";
+
+import * as fs from "fs";
+import { resolve } from "path";
+import * as JSZip from "jszip";
+
+import * as chalk from "chalk";
+
+const program = new Command();
+
+program.name("sb-edit").description("CLI for manipulating Scratch projects");
+
+program
+  .requiredOption("-i, --input <path>", "The path to the input project")
+  .addOption(new Option("-it, --input-type <type>", "The type of input file").choices(["sb3"]))
+  .requiredOption("-o, --output <path>", "The path to the output project")
+  .addOption(new Option("-ot, --output-type <type>", "The type of output file").choices(["leopard", "leopard-zip"]));
+
+program.parse();
+
+const options: {
+  input: string;
+  inputType: "sb3";
+  output: string;
+  outputType: "leopard" | "leopard-zip";
+} = program.opts();
+
+let { input, inputType, output, outputType } = options;
+
+// Infer input type
+if (inputType === undefined) {
+  inputType = "sb3";
+}
+
+// Infer output type
+if (outputType === undefined) {
+  if (output.endsWith(".zip")) {
+    outputType = "leopard-zip";
+  } else {
+    outputType = "leopard";
+  }
+}
+
+async function run() {
+  class StepError extends Error {
+    constructor(message: string) {
+      super(message);
+      this.name = "StepError";
+      Object.setPrototypeOf(this, StepError.prototype);
+    }
+  }
+
+  let stepNumber = 1;
+  async function writeStep<ReturnType>(description: string, fn: () => ReturnType | Promise<ReturnType>) {
+    process.stdout.write(chalk.gray`  ${stepNumber++}. ${description}`);
+    let value: ReturnType;
+    try {
+      value = await fn();
+      process.stdout.write(chalk.bold.green(" Done.\n"));
+    } catch (err) {
+      if (err instanceof StepError) {
+        process.stdout.write(chalk.bold.red(` ${err.message}`));
+        process.stdout.write(chalk.red(`\n\nProject conversion failed.\n`));
+      } else {
+        process.stdout.write(chalk.bold.red` An unknown error occurred.`);
+        process.stderr.write(chalk.red`\n\n${err}\n`);
+      }
+      process.exit(1);
+    }
+    return value;
+  }
+
+  process.stdout.write(chalk.underline.gray`Converting project:\n`);
+
+  const fullInputPath = resolve(process.cwd(), input);
+
+  const project = await writeStep(
+    `${chalk.bold("Importing")} ${inputType} project from path ${chalk.white(fullInputPath)}.`,
+    async () => {
+      let file: Buffer;
+      try {
+        file = fs.readFileSync(fullInputPath);
+      } catch (err) {
+        if (err instanceof Object && "code" in err) {
+          switch (err.code) {
+            case "ENOENT":
+              throw new StepError("File not found.");
+            case "EISDIR":
+              throw new StepError("Input path is a directory but should be a file.");
+          }
+        }
+        throw err;
+      }
+      const project = await Project.fromSb3(file);
+      return project;
+    }
+  );
+
+  switch (outputType) {
+    case "leopard": {
+      const leopard = await writeStep(`${chalk.bold("Converting")} project to ${chalk.white("Leopard")}.`, () => {
+        const leopard = project.toLeopard();
+        return leopard;
+      });
+
+      const fullOutputPath = resolve(process.cwd(), output);
+
+      await writeStep(`${chalk.bold("Exporting")} project to directory ${chalk.white(fullOutputPath)}.`, () => {
+        // First, make sure the output path is an empty directory (create it if it doesn't exist)
+        let existingFiles: string[];
+        try {
+          existingFiles = fs.readdirSync(fullOutputPath);
+        } catch (err) {
+          if (err instanceof Object && "code" in err) {
+            switch (err.code) {
+              case "ENOENT":
+                // Directory does not exist, create it
+                fs.mkdirSync(output, { recursive: true });
+                existingFiles = [];
+                break;
+              case "ENOTDIR":
+                throw new StepError("Output path is a file, not a directory.");
+              default:
+                throw err;
+            }
+          } else {
+            throw err;
+          }
+        }
+
+        if (existingFiles.length > 0) {
+          throw new StepError("Output directory is not empty.");
+        }
+
+        // Write the files
+        for (const [filename, contents] of Object.entries(leopard)) {
+          // Create directories as needed
+          fs.mkdirSync(resolve(fullOutputPath, filename, ".."), { recursive: true });
+
+          // Write the file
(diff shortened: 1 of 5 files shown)


src/cli/index.ts
```diff
@@ -0,0 +1,248 @@
...
+        for (const target of [project.stage, ...project.sprites]) {
+          for (const costume of target.costumes) {
+            const filename = `${target.name}/costumes/${costume.name}.${costume.ext}`;
+            const asset = Buffer.from(costume.asset as ArrayBuffer);
+            fs.mkdirSync(resolve(fullOutputPath, filename, ".."), { recursive: true });
+            fs.writeFileSync(resolve(fullOutputPath, filename), asset);
+          }
+
+          for (const sound of target.sounds) {
+            const filename = `${target.name}/sounds/${sound.name}.${sound.ext}`;
```
towerofnix, line 156: Does it necessarily even make sense to hard-code asset paths, by the way? You're explicitly *not* hard-coding code filenames. Would it be possible, for each target, to get the dirname of its corresponding code filename, then put its `sounds` and `costumes` directories under that directory? I don't know if it's trivial to identify which sprite maps to which code filename though.

I do think this is a good idea beyond just "it's nice to not hard-code things" though, since it would make sure any special character handling that the Leopard converter does is respected and reflected here. It would be awkward if costumes and sounds ended up in a different folder than the code because the CLI doesn't match the logic that `toLeopard` uses.

src/cli/index.ts
```diff
@@ -0,0 +1,248 @@
...
+          // Write the file
+          fs.writeFileSync(resolve(fullOutputPath, filename), contents);
+        }
+
+        for (const target of [project.stage, ...project.sprites]) {
+          for (const costume of target.costumes) {
+            const filename = `${target.name}/costumes/${costume.name}.${costume.ext}`;
+            const asset = Buffer.from(costume.asset as ArrayBuffer);
+            fs.mkdirSync(resolve(fullOutputPath, filename, ".."), { recursive: true });
+            fs.writeFileSync(resolve(fullOutputPath, filename), asset);
```
towerofnix, line 152: You should be using `path.join` instead of hard-coding forward slashes (this might or might not work on Windows). And since you're actually building up the `filename` yourself here, you could store the directory separately and skip the `resolve(..., "..")` / `dirname(...)` call:

```suggestion
            const filedir = path.join(target.name, 'costumes');
            const filename = path.join(filedir, `${costume.name}.${costume.ext}`);
            const asset = Buffer.from(costume.asset as ArrayBuffer);
            fs.mkdirSync(resolve(fullOutputPath, filedir), { recursive: true });
```

src/cli/index.ts
```diff
@@ -0,0 +1,248 @@
...
+        }
+      });
+
+      process.stdout.write(chalk.green`\nProject exported to Leopard format.`);
+      process.stdout.write(chalk.gray` Files written to ${chalk.white(fullOutputPath)}\n\n`);
+      process.stdout.write(
+        chalkBox([
+          { str: "To preview the project, run:" },
+          { str: `$ ${chalk.white`cd ${output}`}`, length: 5 + output.length },
+          { str: `$ ${chalk.white`npx vite`}`, length: 10 }
```
towerofnix, line 170: On a principles level I don't really know about suggesting one arbitrary simple HTTP server over another, but obviously it's nice to have something here, and `npx vite` should generally on the system where the file was generated, at least. You could have some text like "or serve with any HTTP server", but if you don't think it's necessary, that's OK.

src/cli/index.ts
```diff
@@ -0,0 +1,248 @@
...
+}
+
+function chalkBox(lines: { str: string; length?: number }[]): string {
+  const lengths = lines.map(line => line.length ?? line.str.length);
+  const boxInnerWidth = Math.max(...lengths) + 2;
+
+  let outputStr = "";
+  outputStr += chalk.bold.blue`╔${"═".repeat(boxInnerWidth)}╗`;
+  for (const line of lines) {
+    const length = line.length ?? line.str.length;
```
towerofnix, line 231: `chalkBox` is beautiful, thank you.

Can we use [`wcwidth`](https://www.npmjs.com/package/wcwidth) here instead of providing `length`, though? It's a ridiculously popular package for exactly this purpose, and is [an implementation of an IEEE spec](https://github.com/timoxley/wcwidth/blob/master/docs/index.md).

towerofnix (changes requested):
This is really awesome work, thank you!!

We left a bunch of comments, but they're mostly touch-ups for the code, and should be pretty quick to address. We also filed two issues following up this pull request, and did those before reviewing, to get some familiarity with the structure here. Check them out if you're curious: #130, #131

Thanks again! Apart from the code touch-ups, this looks great.

src/cli/index.ts
```diff
@@ -0,0 +1,248 @@
...
+
+const program = new Command();
+
+program.name("sb-edit").description("CLI for manipulating Scratch projects");
+
+program
+  .requiredOption("-i, --input <path>", "The path to the input project")
+  .addOption(new Option("-it, --input-type <type>", "The type of input file").choices(["sb3"]))
+  .requiredOption("-o, --output <path>", "The path to the output project")
+  .addOption(new Option("-ot, --output-type <type>", "The type of output file").choices(["leopard", "leopard-zip"]));
```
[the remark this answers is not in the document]
PullJosh: I wish. The `requiredOption` method does not accept an `Option` object.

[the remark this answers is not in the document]
towerofnix: Can GitHub add a 💔 reaction, please?

src/cli/index.ts
```diff
@@ -0,0 +1,248 @@
...
+  return outputStr;
+}
+
+run().catch(err => {
+  if (err instanceof Error) {
+    console.error(err.stack);
+  }
+
+  process.exit(1);
+});
```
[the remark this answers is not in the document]
PullJosh: When I make this change, I get the following warning:
```
Top-level 'await' expressions are only allowed when the 'module' option is set to 'es2022', 'esnext', 'system', 'node16', 'nodenext', or 'preserve', and the 'target' option is set to 'es2017' or higher.
```

I can fix it by changing `tsconfig.json` to use the following settings:

```diff
{
  "compilerOptions": {
    // ...
-   "target": "es5",
-   "module": "commonjs",
+   "target": "es2017",
+   "module": "esnext",
  }
}
```

My actual understanding of these settings is limited. Is this a breaking change for the sb-edit project?

src/cli/index.ts
PullJosh: Yes, but this feels like a change that should be made to the `.toLeopard()` method. The way assets are currently handled is pretty gross to be honest. It might be nice to round up all the places that this method is being used (such as here and in the Leopard website source code) and examine the usage to design a better API.

src/cli/index.ts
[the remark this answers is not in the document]
towerofnix: Nah, if it's going to involve fiddling with the compiler settings then it isn't worth it within the scope of this pull request. Thanks for looking into this anyway!

src/cli/index.ts
```diff
@@ -0,0 +1,306 @@
...
+          await fs.writeFile(filePath, contents);
+        }
+
+        for (const target of [project.stage, ...project.sprites]) {
+          const costumeDir = path.join(target.name, "costumes");
+          for (const costume of target.costumes) {
+            const filename = path.join(costumeDir, `${costume.name}.${costume.ext}`);
+            const asset = Buffer.from(costume.asset as ArrayBuffer);
+            await fs.mkdir(path.resolve(fullOutputPath, filename, ".."), { recursive: true });
+            await fs.writeFile(path.resolve(fullOutputPath, filename), asset);
```
towerofnix, line 209: This is *almost* exactly on point, but let's move the `fs.mkdir` call out of the for loop here—we only need to create it once for each sprite, not once for each costume in each sprite.

```diff
  const costumeDir = path.join(target.name, "costumes");
+await fs.mkdir(path.resolve(fullOutputPath, costumeDir), { recursive: true });
  for (const costume of target.costumes) {
    const filename = path.join(costumeDir, `${costume.name}.${costume.ext}`);
    const asset = Buffer.from(costume.asset as ArrayBuffer);
-   await fs.mkdir(path.resolve(fullOutputPath, filename, ".."), { recursive: true });
    await fs.writeFile(path.resolve(fullOutputPath, filename), asset);
```

Same thing for sounds below, of course.

towerofnix (changes requested):
Thanks! This all looks good except for one more tweak to how asset directories are created.

src/cli/index.ts
PullJosh: Good call. Thank you

PullJosh:
@towerofnix Fixed the issue you mentioned. I also noticed that the `wcwidth` package was not actually counting `chalk` string widths correctly, so I replaced it with a "stripAnsi" function that does the job. With your approval, I think this is ready to ship.

towerofnix (approved):
Oops, that's a good catch (and close call) with `wcwidth` not doing what I thought it did! Thank you.

And thanks for addressing everything else, too, and for your work on this in general! This looks great. 🥳

PR #137, opened 2024-05-24
toLeopard: Fixes and improved clarity for unique-name behavior

### LINKED ISSUE #133: toLeopard: Sprites named "Math" and "Date" break math and time blocks in their own scripts
If you have a sprite named `Math` or `Date`, its own name will actually overshadow these browser-provided globals! This causes assorted errors when generated code accesses those globals.

### LINKED ISSUE #134: toLeopard: Script names can override certain (newer) important names
These names should be reserved for script names, and currently aren't:

* `"fence"`
* `"nearestEdge"`
* `"positionInFence"`
* `"scratchTan"`
* `"stopAllOfMySounds"` (this is typo'd as `"stapAllOfMySounds"`, LOL)
* `"wrapClamp"`

Creating scripts with these names (or which are these names, when camel-cased) may break project behavior.

### LINKED ISSUE #135: toLeopard: Sprites and the stage shouldn't share the same reserved name list
There's plenty overlap between them, but also certain properties that are only specially meaningful on one of the two!

Sharing the same list doesn't cause any behavior harm, but it's overly cautious and can result in not-nice code. This will be a particularly big deal when variables are stored on `this` rather than `this.vars` (leopard-js/leopard#196).

### LINKED ISSUE #136: toLeopard: Custom block named "constructor" breaks project generation
This actually causes a syntax error, presumably in Prettier or such:

```
SyntaxError: Constructor can't be a generator. (29:11)
  27 |
  28 |         
> 29 |         * constructor() {
     |           ^
  30 |           this.say("Oopsie")
  31 |         }
  32 |
```

Spooky! `"constructor"` needs to be a reserved word.

### LINKED ISSUE #139: toLeopard: "NaN" is a valid sprite name, and a syntax error!
Terrifying! We caught this while glancing over `globals.builtins` for #138. Sprites should definitely never be generated with this name.

towerofnix:
Development:

- Fixes #133 
- Fixes #134 
- Fixes #136
- Fixes #139 
- Resolves #135 
- Towards leopard-js/leopard#196

Addresses the stuff above! Notable details:

* We've moved these lists, still hard-coded, into top-level constants, rather than embedding them right inside the `toLeopard` function. This makes them more convenient to locate and edit, and separates particularly word-based data from actual conversion logic.
* We've updated and added new documentation strings to relevant top-level constants. These hopefully offer some improved clarity — particularly useful since these are important lists to keep maintained!
* We've renamed `uniqueNameGenerator` to `uniqueNameFactory` because generator functions and factory functions are different things, and this is the latter. (Best not to conflate with the JavaScript term "generators", especially since Leopard itself makes extensive use of JS generators.)
* We've improved some internal spot documentation. This probably belongs in the next PR, but it's just better describing the current state of the code, so...

Here's the big external change, besides the word-list fixes listed at the start:

* We're using separate reserved word lists, based on a common list, for the stage and sprites now. We've closely reviewed the current Leopard definitions of `SpriteBase`, `Stage`, and `Sprite` to make sure we're covering everything (and not leaving some crucial things only in the sprite list and not the stage list, for example). But a second pair of eyes comparing that code with this would be great, of course!

We have also introduced a new `JS_RESERVED_PROPERTIES` list. These are names you *really don't* want generated code to ever be messing with! This is mostly to cover our bases when we do variables on `this`. We technically won't ever hit `__proto__` because that's not a camel-case name, and `prototype` is only meaningful on functions (which sprite instances aren't)... but `constructor` really does fix a bug (#136), and we figured better to be a little more comprehensive than less, in such a context-generic list.

### DIFF
--- a/src/io/leopard/toLeopard.ts
+++ b/src/io/leopard/toLeopard.ts
@@ -341,7 +341,7 @@ enum InputShape {
   Stack = "Stack"
 }
 
-function uniqueNameGenerator(reservedNames: string[] | Set<string> = []) {
+function uniqueNameFactory(reservedNames: string[] | Set<string> = []) {
   const usedNames: Set<string> = new Set(reservedNames);
   return uniqueName;
 
@@ -425,7 +425,7 @@ export default function toLeopard(
   let customBlockArgNameMap: Map<Script, { [key: string]: string }> = new Map();
   let variableNameMap: { [id: string]: string } = {}; // ID to unique (Leopard) name
 
-  const uniqueSpriteName = uniqueNameGenerator(LEOPARD_RESERVED_SPRITE_NAMES);
+  const uniqueSpriteName = uniqueNameFactory(LEOPARD_RESERVED_SPRITE_NAMES);
 
   for (const target of [project.stage, ...project.sprites]) {
     const newTargetName = uniqueSpriteName(camelCase(target.name, true));
@@ -437,17 +437,16 @@ export default function toLeopard(
     //
     // Note: since variables are serialized as properties on an object (this.vars),
     // these never conflict with reserved JavaScript words like "class" or "new".
-    let uniqueVariableName = uniqueNameGenerator();
+    const uniqueVariableName = uniqueNameFactory();
 
     for (const { id, name } of [...target.lists, ...target.variables]) {
       const newName = uniqueVariableName(camelCase(name));
       variableNameMap[id] = newName;
     }
 
-    const reservedProperties =
-      target === project.stage ? LEOPARD_RESERVED_STAGE_PROPERTIES : LEOPARD_RESERVED_SPRITE_PROPERTIES;
-
-    const uniqueScriptName = uniqueNameGenerator(reservedProperties);
+    const uniqueScriptName = uniqueNameFactory(
+      target === project.stage ? LEOPARD_RESERVED_STAGE_PROPERTIES : LEOPARD_RESERVED_SPRITE_PROPERTIES
+    );
 
     for (const script of target.scripts) {
       script.setName(uniqueScriptName(camelCase(script.name)));
@@ -457,7 +456,7 @@ export default function toLeopard(
 
       // Parameter names aren't defined on a namespace at all, so must not conflict
       // with JavaScript reserved words.
-      const uniqueParamName = uniqueNameGenerator(JS_RESERVED_WORDS);
+      const uniqueParamName = uniqueNameFactory(JS_RESERVED_WORDS);
 
       for (const block of script.blocks) {
         if (block.opcode === OpCode.procedures_definition) {
--- a/src/io/leopard/toLeopard.ts
+++ b/src/io/leopard/toLeopard.ts
@@ -70,6 +70,35 @@ const JS_RESERVED_WORDS = [
 ];
 
 /**
+ * Global identifiers which Leopard sprites (subclasses of `Sprite`) must not
+ * be named as, since that would overwrite or be a reference binding error.
+ *
+ * Only capitalized identifiers need to be listed here: generated sprite names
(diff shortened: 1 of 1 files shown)


src/io/leopard/toLeopard.ts
```diff
@@ -69,6 +69,215 @@ const JS_RESERVED_WORDS = [
...
+ *
+ * However, it *must* include even (capitalized) identifiers which *aren't*
+ * provided by Leopard, if any part of generated Leopard code could directly
+ * refer to those identifiers (expecting some browser-provided value, rather
+ * than a generated `Sprite` subclass!).
+ */
+const LEOPARD_RESERVED_SPRITE_NAMES = [
+  // Browser-provided identifiers
+  "Date",
+  "Math",
```
PullJosh, line 89: **Question:** Do we want to blacklist browser-provided identifiers like `Array`, `Audio`, `Error`, `File`, `Map`, `Set`, `Temporal`, etc., which are not used in generated Leopard code but could cause conflicts or confusion when the user starts adding their own hand-written JS code?

There are a lot of them. And keeping up-to-date might be challenging.

src/io/leopard/toLeopard.ts
```diff
@@ -69,6 +69,215 @@ const JS_RESERVED_WORDS = [
...
+
+  // Pen
+  "clearPen"
+];
+
+/**
+ * Property names which are used by Leopard sprites (instances of `Sprite`,
+ * whether any subclass or directly constructed from `Sprite`). This list is
+ * a superset of `LEOPARD_RESERVED_SPRITE_BASE_PROPERTIES`.
+ */
```
PullJosh, line 209: This comment is for sprites, but the code is for the stage. The comment below is also reversed.

PullJosh (changes requested):

PullJosh:
This looks nice! Thanks. 😄

I haven't properly tested the behavior, but the code looks good. If you're reasonably confident that the behavior is correct, I'm happy with merging this and then maybe I can give it a proper test of my own after switching away from `this.vars.*` before publishing a release.

Or if you'd rather wait until I can really sit down with it, it'll just take a few days.

towerofnix:
Oops! Thanks for that catch LOL.

We'd prefer that you take the time to check it out before we merge. It fixes some bugs and otherwise is a totally non-breaking release, so we'd kind of like to publish it as a separate version from the change for `this.vars.*` -> `this.*`. Of course we can work out any quirks while testing in general, and the current version will only be useful historically speaking, but it'd still be nice to have this validated and published as its own thing. (The last bastion of `this.vars.*` syntax, lol.)

No problem about the delay, by the way. Thanks to your blessing/look-over we can say this is a good base, so we can get started on `this.vars.*` -> `this.*` in a separate branch sooner, just starting from these commits.

src/io/leopard/toLeopard.ts
towerofnix: Thanks, this is a good point! We've followed it up in detail here: #138

towerofnix:
Addendum for this PR: While we can [generate the lists dynamically](https://github.com/leopard-js/sb-edit/issues/138#issuecomment-2132171014) and certainly [include a broader list of globals](https://github.com/leopard-js/sb-edit/issues/138#issuecomment-2131898569) than just what sb-edit depends on accessing—we are still OK to start on implementations based on these top-level constants. Because, we'd just be changing how the lists are generated, not how they are externally accessed (still the same variable names that still represent the same practical contents, just better versions of those lists).

PullJosh (approved):

PullJosh:
Tested and looks good! 😄 (Sorry for the delay)

towerofnix:
No problem! We've been busy anyway LOL. 🥳

PR #142, opened 2024-05-30
Add -t, --trace CLI option; enable TypeScript source maps

towerofnix:
This PR adds a `-t` / `--trace` CLI option which enables showing the log for unknown errors. It's not necessarily meant to be used by users, moreso just a convenience for us debugging problems.

It also enables source map generation in tsconfig.json. I believe these would be included in the NPM package, which I think is standard practice. Node.js (when running sb-edit) doesn't actually do anything with source maps without providing the `--enable-source-maps` option, typically via an environment variable.

sb-edit output with Node.js --enable-source-maps and sb-edit -t. The message reads, An unknown error occurred. The JavaScript details are shown below. This is a bug, please share the project you used on: URL to sb-edit issues. Then there is a blank line and the full error trace, which includes paths to the TypeScript source files.

Same as above but no --enable-source-maps. Everything is the same except you get generated lib files in the stack trace, instead of source TypeScript files.

Typical sb-edit output without any special options. The message is the same and you still see the JavaScript description of the error, but no stack trace.

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -16,7 +16,8 @@ program
   .requiredOption("-i, --input <path>", "The path to the input project")
   .addOption(new Option("-it, --input-type <type>", "The type of input file").choices(["sb3"]))
   .requiredOption("-o, --output <path>", "The path to the output project")
-  .addOption(new Option("-ot, --output-type <type>", "The type of output file").choices(["leopard", "leopard-zip"]));
+  .addOption(new Option("-ot, --output-type <type>", "The type of output file").choices(["leopard", "leopard-zip"]))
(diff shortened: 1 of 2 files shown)


towerofnix:
Uh, we can't read and we clearly did not get source mapped stack traces working. 😅 One moment!

towerofnix:
OK, we switched to inline source maps and apparent Node.js likes those more. I'm sure separate-file source maps are supported in some manner but I can't really find documentation on it and am not too worried about it ATM (the codebase is small, I don't think this adds any remotely significant JS parser/memory overhead).

Same output but with real TypeScript traces now

PullJosh (approved):
This looks good! Thank you.

Issue #141, opened 2024-05-30
Cannot read properties of undefined (reading 'opcode')


Administrator95:
https://scratch.mit.edu/projects/1025698237/
I was trying to run this project on the website and in NodeJS. However, I keep receiving this error. The only extension it uses is pen and it runs on Scratch 3.0.

towerofnix:
Thanks for reporting this! We've looked into the things that kept this project from loading and have figured out fixes in pull requests #145 and #146. Once those are both merged, the project should load correctly.

Unfortunately, it doesn't seem to be totally operational. The project runs and the HTML parser appears to be working correctly (probably?) but nothing ends up showing on the screen, except for the grey background. I don't know for sure, but I think the most likely cause is the pen text engine may be using special-characters in its costumes that sb-edit and Leopard don't handle super nicely yet. We are tracking this in #140 but haven't coded a fix yet. If you could write a simple script that tests the behavior of PixelMplus on its own, and doesn't interact with any other modules in your code, that would probably help tell for sure if the text engine is having trouble or if it's something else. (It could be both... But one thing at a time, with debugging... ^^;)

PR #145, opened 2024-05-30
fromSb3: Fix loading component-based pen color blocks; document shadow blocks' fields more carefully

### LINKED ISSUE #144: fromSb3: Peculiarity with pen_menu_colorParam as shadow input (etc?)
Investigating #141.

The basic processing for shadow inputs is to copy the inputs and fields of a shadow input into the parent block:

```js
// In most cases, just copy the shadow block's fields and inputs
// into its parent
result = {
  ...result,
  ...translateInputs(blocks[value].inputs),
  ...translateFields(blocks[value].fields, blocks[value].opcode)
};
```

This is usually good enough, but there can be discrepancies between the names of inputs/fields on a shadow input and inputs/fields on the parent block. These need to be handled manually and aren't detected by TypeScript.

One such came up with `pen_menu_colorParam`, which has a strangely camelCased field:

```js
export const fieldTypeMap = {
  ...
  [OpCode.pen_menu_colorParam]: { colorParam: "penColorParam" },
}
```

`pen_menu_colorParam` is a CONSTANT_CASED input on two blocks:

```js
export const inputPrimitiveOrShadowMap = {
  ...
  [OpCode.pen_changePenColorParamBy]: { COLOR_PARAM: OpCode.pen_menu_colorParam, VALUE: BIS.MATH_NUM_PRIMITIVE },
  [OpCode.pen_setPenColorParamTo]: { COLOR_PARAM: OpCode.pen_menu_colorParam, VALUE: BIS.MATH_NUM_PRIMITIVE },
};
```

This results in `COLOR_PARAM` being undefined (possibly only in certain project.json scenarios), which breaks output formats that expect it to be there. Also a surprise `colorParam` input existing, of course, which isn't referenced by any output formats (it's not part of the sb-edit intermediate representation).

It would be nice if TypeScript could flag cases like this automatically. Right now `inputPrimitiveOrShadowMap` isn't typed at all, so it looks like some work could be done here.

towerofnix:
Development:

- Resolves #144 
- Towards #141 (specific project)

Hard-codes(!) a special-case fix for the `pen_menu_colorParam` shadow block; we reviewed all shadow blocks that Leopard supports and would have included like fixes for any similar cases, but this is the only one.

`pen_menu_colorParam` is different from normal SB3 shadow blocks because its field is named `colorParam` and does not match the name, `COLOR_PARAM`, of the input into which it is placed. All other names *do* match and we can just apply the shadow block's field as an sb-edit input with the same name, but this one needs to be manually transferred from sb3 field `colorParam` to sb-edit input `COLOR_PARAM`.

This fixes loading blocks which use this input (there are two in the Pen extension) and makes it so output formats see the `COLOR_PARAM` input that they're expecting. Our testing:

* Manually tested with project [Set/change pen (color...) menu tests](https://scratch.mit.edu/projects/1029845739/)
* Manually tested with project [WebOMatic!](https://scratch.mit.edu/projects/1025698237/)
* Both `toLeopard` and `toSb3` are working, and don't require any changes.
    * Included reviewing generated project.json for first project, see: https://github.com/leopard-js/sb-edit/issues/144#issuecomment-2140119310
    * Not directly related, but Leopard's component-based pen color does work with these blocks correctly!
* Updated snapshot test (which loads with `fromSb3` and writes with  `toLeopard`,`toSb3`, and `toScratchblocks`)
    * Internal `project.json` is now pretty-printed w/ `JSON.stringify` for more convenient git review. The SB3 is very old and saving it from the Scratch editor *without* making any changes will still produce a flat-out different file (in subtle ways), so the diff was manually reviewed to only commit the code changes. This should make for a clean diff in the snapshot file.


towerofnix:
Have requested a review from @adroitwhiz since this is probably at least of note to you, but all cool if you aren't around to look at this closely at the moment!

adroitwhiz (approved):
LGTM! At some point, it may be worth revisiting shadow blocks but this should fix things for now.

PullJosh (approved):

PR #146, opened 2024-05-30
fromSb3: Fix loading undefined sb3Block.next

### LINKED ISSUE #143: fromSb3: Peculiarities in blockWithNext
Investigating #141.

A simple change in `blockWithNext` to help debug it:

```diff
 if (sb3Block.next !== null) {
+  console.log(block.opcode, '->', sb3Block.next);
   next = blockWithNext(sb3Block.next, blockId);
 }
```

Last few messages leading up to error are:

```
data_deletealloflist -> c:
data_deletealloflist -> F
procedures_call -> %
procedures_call -> mk
argument_reporter_string_number -> undefined
```

This project doesn't appear to be generated from a normal Scratch editor, i.e. those are extremely short IDs compared to a normal project:

```
_whenflagclicked -> qSAt|FhG1SUUsVRn{iDY
procedures_definition -> :r0{L0I?SV?{-Hk.a(.D
event_whenflagclicked -> qLuxW0/!F{U*KyZ0T6tJ
```

So it's quite possible that the `sb3Block.next === undefined` instead of `null` like we're checking for is an exemption the minifier or project saver this project was made with took. (Unmodified scratch-vm has no problem loading this.)

Still, it seems like this is something we are more or less trying to detect but... not treating appropriately...? Check out a little bit earlier in `blockWithNext`:

```js
const block = new BlockBase({
  opcode: sb3Block.opcode,
  inputs: getBlockInputs(sb3Block, blockId),
  id: blockId,
  parent: parentId,
  next: sb3Block.next ?? undefined
}) as Block;
```

What's with that `?? undefined`, if we're going to be comparing it against `null` immediately thereafter? Shouldn't it be `?? null`?

towerofnix:
Development:

- Resolves #143 
- Towards #141 (specific project)

See #143 for some technical discussion.

Best to view the commits separately, as they are independent changes, though they affect the same area of code.

* Makes undefined (missing) `next` on blocks in a project.json not crash `fromSb3`. This may be a omission by a minifier used in the demo project, rather than something Scratch itself will ever serialize, though Scratch handles deserializing undefined `next` just fine, so sb-edit should too.
* Improves some code clarity. `blockWithNext` is sometimes called without its `parentId` argument so we just make it `parentId?: string` instead of `parentId: string | undefined`. And `getBlockScript`, previously called once for each script in a target, is now called just once for each target, because `getBlockScript` does not care which script you're about to get, and is structured to make all scripts available. (It's a currying function, so now we use it like one!)

Tested manually with the project shared in #141, [WebOMatic!](https://scratch.mit.edu/projects/1025698237/).

### DIFF
--- a/src/io/sb3/fromSb3.ts
+++ b/src/io/sb3/fromSb3.ts
@@ -292,7 +292,7 @@ function getBlockScript(blocks: { [key: string]: sb3.Block }) {
     }
   }
 
-  function blockWithNext(blockId: string, parentId: string | undefined = undefined): Block[] {
+  function blockWithNext(blockId: string, parentId?: string): Block[] {
     const sb3Block = blocks[blockId];
     const block = new BlockBase({
       opcode: sb3Block.opcode,
@@ -302,7 +302,7 @@ function getBlockScript(blocks: { [key: string]: sb3.Block }) {
       next: sb3Block.next ?? undefined
     }) as Block;
     let next: Block[] = [];
-    if (sb3Block.next !== null) {
+    if (typeof sb3Block.next === "string") {
       next = blockWithNext(sb3Block.next, blockId);
     }
     return [block, ...next];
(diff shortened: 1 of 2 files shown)


PullJosh (approved):
Tested and looks good 👍 Thank you!

PR #147, opened 2024-06-25
Re-save test.sb3 from current Scratch editor, update snapshot

towerofnix:
The only in-project edits we made were to 1) manually show the list, because the old format appears to not make the monitor visible, mismatching the snapshot, and to 2) add an "inert" custom block which just runs "stop all", and put that at the top of every script, because otherwise *doing practically anything, including opening the project,* would mutate the project. (Most specifically, by hiding the list in response to an effective "when timer > 0" block.)

This doesn't implement new tests or make other changes to testing — it just brings the sb3 up to date with current Scratch standards, so it can be re-saved without grievances from the snapshot test. (We did this and ran `npm run test` again just to make sure, but when reviewing, you should, too!)


towerofnix:
Note re: snapshot changes in #145, we didn't pretty-print the JSON for test.sb3 this time. It's just a straight export from the Scratch editor. Since future changes should not require any manual JSON finagling, it doesn't save a step for diff convenience anymore, really.

PullJosh (approved):
Looks good! 👍 Thank you for the cleanup.

PR #148, opened 2024-07-01
Add "--leopard-url" CLI option

towerofnix:
Enables testing a custom Leopard build locally, or providing any other alternate Leopard library host.

The default value is the same one that is used by default in [`toLeopard.ts`](https://github.com/leopard-js/sb-edit/blob/9b0f978422d8c0c55918267f7721ff00cb575453/src/io/leopard/toLeopard.ts#L405-L406). It would be nice not to duplicate this definition, but unless toLeopard defined it as an export (this would probably be awkward) or stops providing a default at all, it is what it is.

This option is used in both leopard and leopard-zip output formats, and ignored in other formats.

Only fully qualified URLs are supported, i.e. you can't just do `/leopard/dist`. This is to make the output folder work regardless where it's hosted (provided the specified URL is accessible there), and to avoid confusing a path on a hypothetical file server with a path on your own system.

There's no magic about e.g. taking a specified *path* to Leopard and copying out the dist/ files at time of `sb-edit` run (to make a self-contained folder/zip). That's out of scope for this PR, and may require adjustments to `toLeopard` internals (e.g. supporting relative imports there).

Typical development setup 1:

```sh
# terminal 1
cd leopard
npm run dev

# terminal 2
cd leopard/dist
npx vite --port 6201

# terminal 3
npx vite --port 6200

# terminal 4
sb-edit -i /path/to/proj.sb3 -o demo \
  --leopard-url http://localhost:6201/

# in browser, navigate to:
# http://localhost:6200/demo/
```

Typical development setup 2:

```sh
# terminal 1
cd leopard
npm run dev

# terminal 2
npx vite --port 6200

# terminal 3
rm -rf demo
sb-edit -i /path/to/proj.sb3 -o demo \
  --leopard-url http://localhost:6200/leopard/dist/

# in browser, navigate to:
# http://localhost:6200/demo/
```

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -17,7 +17,10 @@ program
   .addOption(new Option("-it, --input-type <type>", "The type of input file").choices(["sb3"]))
   .requiredOption("-o, --output <path>", "The path to the output project")
   .addOption(new Option("-ot, --output-type <type>", "The type of output file").choices(["leopard", "leopard-zip"]))
-  .addOption(new Option("-t, --trace", "Show a detailed error trace"));
+  .addOption(new Option("-t, --trace", "Show a detailed error trace"))
+  .addOption(
+    new Option("--leopard-url <url>", "The URL to use for Leopard").default("https://unpkg.com/leopard@^1/dist/")
+  );
 
 program.parse();
 
@@ -27,6 +30,7 @@ const options: {
   output: string;
   outputType: "leopard" | "leopard-zip";
   trace: boolean | undefined;
+  leopardUrl: string;
 } = program.opts();
 
 let { input, inputType, output, outputType } = options;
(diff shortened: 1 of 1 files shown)


src/cli/index.ts
```diff
@@ -163,11 +167,30 @@ async function run() {
     }
   );
 
+  function toLeopard() {
+    const { leopardUrl: leopardURL } = options;
```
PullJosh, line 171: Not particularly important, but is there any reason why this isn't simplified?

```diff
- const { leopardUrl: leopardURL } = options;
+ const { leopardUrl } = options;
```

PullJosh (commented):
Aside from one tiny comment, I love it!

src/cli/index.ts
PullJosh: Oh, yes! I didn't even recognize that there was a rename happening here. I read it (incorrectly) as `{ leopardUrl: leopardUrl }`.

PullJosh (approved):

towerofnix:
Merged, thanks! I'm going to push 0.15.0 with this and other recent merges.

PR #149, opened 2024-07-01
toLeopard: Strict number casting

### LINKED ISSUE #123: toLeopard: Static number casting issues in equality operator + NaN leaking

towerofnix:
Resolves #123. Related to leopard-js/leopard-mentors#4 and #150.

This PR implements the proposal most particularly outlined in https://github.com/leopard-js/sb-edit/issues/123#issuecomment-1868418197. It's intended as an alternative, and probably lesser, option, compared to the "traits" system outlined in - well, the same comment ✨ - and we'd like to have a go at implementing that one, before merging this. However, both PRs should be equally effective in generally fixing issues to do with surprise NaN in converted projects. (That being the case, behavior-testing projects are much needed.)

We haven't looked into it in great detail yet, but it appears to - more or less - fix the behavior of the case project (leopard-js/leopard-mentors#4), which has code that apparently sets sprite sizes to NaN.

This code is from December 2023 and we haven't made any significant edits except for rebasing it onto current master.

### DIFF
--- a/src/__tests__/__snapshots__/compilesb3.test.ts.snap
+++ b/src/__tests__/__snapshots__/compilesb3.test.ts.snap
@@ -266,7 +266,7 @@ export default class Stage extends StageBase {
       "thing"
     );
     this.vars.testList.splice(
-      this.itemOf(this.vars.testList, 0) - 1,
+      this.toNumber(this.itemOf(this.vars.testList, 0)) - 1,
       1,
       this.vars.testList.length
     );
--- a/src/io/leopard/toLeopard.ts
+++ b/src/io/leopard/toLeopard.ts
@@ -298,11 +298,32 @@ enum InputShape {
   Any = "Any",
 
   /**
-   * Number input shape. If the input block isn't guaranteed to be a number,
-   * it is automatically wrapped with this.toNumber(), which has particular
-   * behavior to match Scratch.
+   * Typical number input shape. If the input block isn't guaranteed to be
+   * a number, it is automatically wrapped with this.toNumber(), which has
+   * particular behavior to match Scratch: most values that JavaScript casts
+   * to NaN are instead cast to zero.
    */
-  Number = "Number",
+  ZeroCastNumber = "ZeroCastNumber",
+
+  /**
+   * Number input shape which requires that non-numeric values get cast to NaN
+   * instead of zero. This is mostly useful as a desired input shape for equals
+   * comparison, but some blocks also satisfy it directly. While ZeroCastNumber
+   * satisfies StrictlyCastNumber, the reverse isn't true - blocks which return
+   * a StrictlyCastNumber will be cast when ZeroCastNumber is desired, avoiding
+   * NaN "leaking" into numeric operations that expect zero instead.
+   */
+  StrictlyCastNumber = "StrictlyCastNumber",
+
+  /**
+   * Number input shape where the dynamic behavior is the same regardless if
+   * the input is equal to zero or NaN. This avoids needless casting from
+   * StrictlyCastNumber to ZeroCastNumber, and uses the more "simple" casting
+   * of this.toNumber() when casting is required (i.e. the reporter doesn't
+   * return a number at all). Mostly useful for duration or quantity-based
+   * operations.
+   */
+  LooseNumber = "LooseNumber",
 
   /**
    * Special "index" shape, representing an arbitrary number which has been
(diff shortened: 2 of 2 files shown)


towerofnix:
*Note for reviewing: refer to https://github.com/leopard-js/sb-edit/pull/150#issuecomment-2204039969*

Issue #100, opened 2023-03-06
Block inputs' types don't appear to ever account for reporter blocks being plugged in


adroitwhiz:
I really hope I'm just missing something here, because this seems to undermine the entire type-safety of the library.

Right now, it seems that the `Block` type is set up such that a block's inputs are always assumed to hold literal values and never any reporter blocks plugged into them. For instance, the `operators_add` block is always assumed to have two `BlockInput.Number` inputs, each of which holds a numeric literal. The possibility that these inputs may *also* be occupied by other (reporter) blocks doesn't seem to be considered.

So far, there are a couple reasons that this has managed to sneak through the API:
- All serialization code written so far seems to have a generic "serialize inputs" function that treats *all* inputs as `BlockInput.Any`, meaning that it handles the case where those inputs are blocks.
- The `Block` type is a union of `KnownBlock` and `UnknownBlock`. Since `UnknownBlock` is actually just a supertype of `KnownBlock` where the opcode can be any string (including redefining known opcodes) and all inputs are `BlockInput.Any`, all code which uses the `Block` type cannot make any assumptions about the "shape" of the block's inputs based on its opcode, and must handle all cases.

Even so, I've already found a case where serialization errors out as a result of incorrectly narrowing blocks' inputs based on opcodes: the "change pen (param) by" block cannot be converted if a reporter block is plugged into the menu:
![image](https://user-images.githubusercontent.com/25993062/223012964-4daeb4c5-5d9a-4be4-aa61-8714a5d3c85d.png)
![image](https://user-images.githubusercontent.com/25993062/223013067-6c587e36-e5c8-4e79-8b6d-f20e3c08465e.png)

There are probably many more instances of this throughout the codebase. I think there needs to be a major rework of the `Block` class in order to fix this soundness hole, especially if we want to use TypeScript to its full potential.

towerofnix:
I'd say you're (unfortunately!) right in your assessment here — I ran into a number of issues which were flavored by this when reworking `toLeopard.ts` recently.

> We may want to consider separating inputs and fields the same way Scratch does, so that we can tell which inputs may and may not have blocks plugged into them.

I'm feeling this. Since Scratch 3.0 is the base for `KnownBlock`, we should be able to accurately and completely reflect information about those blocks in sb-edit — including the difference between inputs and fields.

[...]

PR #154, opened 2024-07-09
Bump most dev dependencies

adroitwhiz:
Bumps eslint, prettier, and typescript, along with associated configs. ESLint has a new config format, which I've migrated to.

Newer versions of typescript-eslint and TypeScript seem to have more lints--I've disabled some and manually ignored others.

Prettier v3 is async, so we can't use it in toLeopard. Instead, we continue using v2 there but use v3 for the codebase itself.


src/io/leopard/toLeopard.ts
```diff
@@ -4,7 +4,9 @@ import Block from "../../Block";
 import * as BlockInput from "../../BlockInput";
 import { OpCode } from "../../OpCode";
 
-import * as prettier from "prettier";
+// TODO: Upgrade to Prettier v3, which formats asynchronously
+// Or https://github.com/prettier/prettier-synchronized if it's OK to be Node-only
+import * as prettier from "prettier2";
```
towerofnix, line 9: prettier-synchronized sounds a little evil (with due respect!). It makes sense not to bother with prettier 2->3 (at runtime) in this PR, but is this something we could follow up on pretty quickly? 

The only frontends that I believe actually interface with `toLeopard` are sb-edit's newish CLI (#129) and leopard-website's [`exportProjectWithBufferAssetsToZip`](https://github.com/PullJosh/leopard-website/blob/b6d56ef7d8adc35f83c603eaf6ef77a9b62fa1c5/lib/convertProject.ts#L37), which is trivial to async-ify (it is already evaluated only in async contexts).

src/io/sb3/fromSb3.ts
```diff
@@ -161,9 +161,10 @@ function getBlockScript(blocks: { [key: string]: sb3.Block }) {
               const shadowInputs = translateInputs(blocks[value].inputs);
               const shadowFields = translateFields(blocks[value].fields, blocks[value].opcode);
 
-              if (blocks[value].opcode === "pen_menu_colorParam") {
+              if (blocks[value].opcode === OpCode.pen_menu_colorParam) {
                 result = {
                   ...result,
+                  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
                   COLOR_PARAM: (shadowFields as any).colorParam
```
towerofnix, line 168: Seems evil, obviously, but I'm OK with not addressing existing logic/safety issues as part of this PR. This instance sounds related to #100. Is it surprising or expected, that this is basically the only related line that new TypeScript checks caught?

towerofnix (approved):
Basically looks great, just a couple of comments. Thank you!

src/io/sb3/fromSb3.ts
adroitwhiz: I believe so, yes.

src/io/leopard/toLeopard.ts
adroitwhiz: I can create a PR to leopard-website, and I'll update to prettier v3 here after that goes in

PullJosh:
Once this is merged, perhaps it's time for an sb-edit release on npm?

towerofnix:
Would in effect involve publishing #122 (affects users), #147 (no affect on users), and this PR ("no" affect on users but good for maintenance). Means saving #150 (affects users) til next - no problem here. This set sounds reasonable to us - good to have compatibility fixes available sooner than later.

Filed #156 for some discussion about a 1.0.0 release in particular, somewhat related to these changes.

PR #157, opened 2024-07-12
Add in support for exporting as a Patch file

benmontycomputer:
This adds support for exporting as a Patch project file (https://codepatch.org).


PullJosh:
I'm so excited to dive into this! 😄 sb-edit support for Patch is so cool.

It will take time to look at and test the code, but I noticed from a quick scan that this PR includes the file `.vscode/settings.json`. If you could remove that file from the PR that would be lovely. Thanks!

benmontycomputer:
My bad, I've cleaned that up.

PR #159, opened 2025-12-27
fix(toLeopard): cast switch costume input to string

qindongliang:
## Description
This PR fixes a bug where `switch costume to [0]` blocks were incorrectly transpiled as `this.costume = 0` (integer), causing Leopard to interpret it as a **costume index** rather than a **costume name**.

In Leopard (and Scratch):
- **Integer `0`**: Interpreted as a 1-based index with wrapping, which resolves to the **last costume** in the list.
- **String `"0"`**: Interpreted as a costume name, correctly finding the costume named "0".

By changing the `InputShape` for `looks_switchcostumeto` from `Any` to `String`, the transpiler now generates `this.costume = "0"`, ensuring the intended costume is selected by name.

## Issue Fixed
Fixes an issue where sprites would unexpectedly switch to the last costume (often causing visual artifacts like "stretching" if the last costume is a specific animation frame) when the intention was to switch to the first costume named "0".

## Verification
Tested with a Scratch project ("Sprunked") that uses a costume named "0".
- **Before**: Transpiled code used `this.costume = 0`, causing the sprite to render the wrong costume (index 16).
- **After**: Transpiled code uses `this.costume = "0"`, correctly rendering the first costume.

### DIFF
--- a/.agent/workflows/convert-scratch-project.md
+++ b/.agent/workflows/convert-scratch-project.md
@@ -1,0 +1,42 @@
+---
+description: 将指定的 .sb3 文件转换为 Leopard 项目并自动关联本地库
+---
+
+此工作流执行从 Scratch (`.sb3`) 到 Leopard (JavaScript) 的全自动化转换，并自动链接本地修复后的 Leopard 框架。
+
+### 参数说明
+- **SB3_PATH**: `.sb3` 文件的绝对路径（例如 `/Downloads/my_game.sb3`）
+- **OUTPUT_DIR**: 转换后项目的保存父目录（默认为 `/tmp`）
+
+---
+
+### 执行步骤
+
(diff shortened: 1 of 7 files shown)


adroitwhiz:
It looks like there are a bunch of other unrelated changes in your `master` branch. Can you create a new branch, and use `git cherry-pick` to select *just* the fix?

qindongliang:
> It looks like there are a bunch of other unrelated changes in your `master` branch. Can you create a new branch, and use `git cherry-pick` to select _just_ the fix?

Sorry for the mix-up. I didn’t notice the unrelated changes in master. I’ll cherry-pick just the fix, create a new branch, and submit a separate PR soon. Thanks for the heads-up!

PR #160, opened 2026-01-14
fix(toLeopard): cast switch costume input to string

qindongliang:
## Description
This PR fixes a bug where `switch costume to [0]` blocks were incorrectly transpiled as `this.costume = 0` (integer), causing Leopard to interpret it as a **costume index** rather than a **costume name**.

In Leopard (and Scratch):
- **Integer `0`**: Interpreted as a 1-based index with wrapping, which resolves to the **last costume** in the list.
- **String `"0"`**: Interpreted as a costume name, correctly finding the costume named "0".

By changing the `InputShape` for `looks_switchcostumeto` from `Any` to `String`, the transpiler now generates `this.costume = "0"`, ensuring the intended costume is selected by name.

## Issue Fixed
Fixes an issue where sprites would unexpectedly switch to the last costume (often causing visual artifacts like "stretching" if the last costume is a specific animation frame) when the intention was to switch to the first costume named "0".

## Verification
Tested with a Scratch project ("Sprunked") that uses a costume named "0".
- **Before**: Transpiled code used `this.costume = 0`, causing the sprite to render the wrong costume (index 16).
- **After**: Transpiled code uses `this.costume = "0"`, correctly rendering the first costume.

### DIFF
--- a/src/io/leopard/toLeopard.ts
+++ b/src/io/leopard/toLeopard.ts
@@ -583,9 +583,9 @@ export default function toLeopard(
     if (script.hat && script.hat.opcode === OpCode.procedures_definition) {
       return `
         * ${script.name}(${script.hat.inputs.ARGUMENTS.value
-        .filter(arg => arg.type !== "label")
-        .map(arg => arg.name)
-        .join(", ")}) {
+          .filter(arg => arg.type !== "label")
+          .map(arg => arg.name)
+          .join(", ")}) {
           ${body}
         }
       `;
@@ -1034,7 +1034,7 @@ export default function toLeopard(
         case OpCode.looks_switchcostumeto: {
           satisfiesInputShape = InputShape.Stack;
 
-          const costume = inputToJS(block.inputs.COSTUME, InputShape.Any);
+          const costume = inputToJS(block.inputs.COSTUME, InputShape.String);
           blockSource = `this.costume = ${costume}`;
 
           break;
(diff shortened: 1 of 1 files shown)


adroitwhiz:
Are you sure your fix is correct? [Scratch itself always treats numbers as costume indices.](https://github.com/scratchfoundation/scratch-editor/blob/1d06762b34483eaa0d88d633d7146df85f1c9229/packages/scratch-vm/src/blocks/scratch3_looks.js#L390-L393)

I believe the actual problem here is that `staticBlockInputToLiteral` [casts `InputShape.Any` inputs to numbers if possible](https://github.com/leopard-js/sb-edit/blob/8e26c36367fa18f391df7332629477278ad82226/src/io/leopard/toLeopard.ts#L518-L527), which seems plainly incorrect. The [documentation for `InputShape.Any`](https://github.com/leopard-js/sb-edit/blob/8e26c36367fa18f391df7332629477278ad82226/src/io/leopard/toLeopard.ts#L294-L298) clearly says that "The input will never be cast", yet `staticBlockInputToLiteral` casts it to a number anyway. I'll create a PR removing the cast.

qindongliang:
> Are you sure your fix is correct? [Scratch itself always treats numbers as costume indices.](https://github.com/scratchfoundation/scratch-editor/blob/1d06762b34483eaa0d88d633d7146df85f1c9229/packages/scratch-vm/src/blocks/scratch3_looks.js#L390-L393)
> 
> I believe the actual problem here is that `staticBlockInputToLiteral` [casts `InputShape.Any` inputs to numbers if possible](https://github.com/leopard-js/sb-edit/blob/8e26c36367fa18f391df7332629477278ad82226/src/io/leopard/toLeopard.ts#L518-L527), which seems plainly incorrect. The [documentation for `InputShape.Any`](https://github.com/leopard-js/sb-edit/blob/8e26c36367fa18f391df7332629477278ad82226/src/io/leopard/toLeopard.ts#L294-L298) clearly says that "The input will never be cast", yet `staticBlockInputToLiteral` casts it to a number anyway. I'll create a PR removing the cast.

Good point! I totally agree with your analysis. The behavior of staticBlockInputToLiteral casting InputShape.Any inputs to numbers does contradict the documentation stating that such inputs "will never be cast".
I’ll go ahead and close my PR right away, and I’m looking forward to your PR that removes this incorrect casting logic.

